Issue #7, opened 2022-05-27
[Feature] FeatureFlagConfiguration should track the related consumers of the CR


AlexsJones:
In order to keep a listing of who is consuming the feature flag CR then we should implement a track in the status field 
e.g. 
```
status: 
  consumers:
  - name:
    namespace:
    type:
```

AlexsJones:
We have decided this is not an ideal pattern and instead will maintain a reference annotation from the configmap to the custom resource

thschue:
👍 

Will close this issue

PR #13, opened 2022-05-29
Add rudimentary reconciliation fn

toddbaert:
I implemented a very rudimentary reconciliation function. I'm sure this isn't production quality.

I tried to avoid the looping, but I wasn't able to get the ListOption field matchers to do the filtering I needed.

Anyway, if a FeatureFlagConfiguration CR is modified now, this updates the associated configMaps, though they take 10-20 seconds to appear updated to the container. Added lots of comments/thoughts inline.

### DIFF
--- a/controllers/featureflagconfiguration_controller.go
+++ b/controllers/featureflagconfiguration_controller.go
@@ -18,7 +18,9 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


AlexsJones:
Thanks @toddbaert might be worth talking to @thschue who is going to add a status field on the Custom Resource to associate known configmaps that are leveraging the flags as per this issue https://github.com/open-feature/open-feature-operator/issues/7 it might make this logic a bit easier to implement.

controllers/featureflagconfiguration_controller.go
```diff
@@ -47,10 +49,40 @@ type FeatureFlagConfigurationReconciler struct {
...
+	// I'd like to be able to do some field-matching with the ListOptions so we don't have to iterate,
+	// but they can't seem to be used for annotations.
+	// we may want our webhook to also add labels for easy querying?
+	if err := r.List(ctx, podList); err != nil {
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// find any pods that are associated with this CR
+	for _, pod := range podList.Items {
+		if (pod.ObjectMeta.Annotations["openfeature.dev/featureflagconfiguration"] == req.Name) {
```
AlexsJones, line 65:
```suggestion
if val, ok := pod.ObjectMeta.Annotations["openfeature.dev/featureflagconfiguration"]; ok {
if val == req.Name {
```

toddbaert:
> Thanks @toddbaert might be worth talking to @thschue who is going to add a status field on the Custom Resource to associate known configmaps that are leveraging the flags as per this issue #7 it might make this logic a bit easier to implement.

Interesting. Is storing such state in the CR idiomatic? It occurs to me that we also could perhaps, in the mutating admission webhook, add a label that might allow easy queries by adding a label matcher to the ListOptions and querying affected for configmaps that way.

AlexsJones:
> > Thanks @toddbaert might be worth talking to @thschue who is going to add a status field on the Custom Resource to associate known configmaps that are leveraging the flags as per this issue #7 it might make this logic a bit easier to implement.
> 
> Interesting. Is storing such state in the CR idiomatic? It occurs to me that we also could perhaps, in the mutating admission webhook, add a label that might allow easy queries by adding a label matcher to the ListOptions and querying affected for configmaps that way.

The state in the CR might become inflated over time that's true, we might even need to create another object to express the relation between the FeatureFlagConfiguration and the ConfigMap that have been created from it.  https://github.com/open-feature/open-feature-operator/issues/14

To find ConfigMaps created by the `MutatingAdmissionWebhook` I have added this ticket to make your life hopefully a lot easier

controllers/featureflagconfiguration_controller.go
```diff
@@ -47,10 +49,41 @@ type FeatureFlagConfigurationReconciler struct {
...
+if val == req.Name {
+			configMapList := &v1.ConfigMapList{}
+			
+			// query configMaps
+			r.List(ctx, configMapList) 
+			for _, configMap := range configMapList.Items {
+				// find the configMap matching the pod name (this is how our webhook names them for now, might want something else long-term)
+				if (configMap.Name == pod.Name) {
+
+					// get the new contents by querying our CR based on the request data.
```
AlexsJones, line 75: We need a way of actually comparing the contents of the configmap config.yaml against the FeatureFlagConfiguration Spec internals

toddbaert: Created: https://github.com/open-feature/open-feature-operator/issues/20

AlexsJones:
Closing this @toddbaert as per our conversation, this was a good first start, but will need to be updated against the new reconciliation logic if you want to take it further!

PR #18, opened 2022-05-30
feat: change configmap creation behavior

thschue:


webhooks/mutating_admission_webhook.go
```diff
@@ -119,20 +160,10 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
 }
 
-// PodMutator implements admission.DecoderInjector.
-// A decoder will be automatically injected.
-
-// InjectDecoder injects the decoder.
-func (m *PodMutator) InjectDecoder(d *admission.Decoder) error {
-	m.decoder = d
-	return nil
+func False() *bool {
```
AlexsJones, line 166: This can be private

webhooks/mutating_admission_webhook.go
```diff
@@ -42,66 +42,107 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+		reference := pod.OwnerReferences[0]
+		reference.Controller = False()
+		cm.OwnerReferences = append(cm.OwnerReferences, reference)
+		err := m.Client.Update(ctx, &cm)
+		if err != nil {
+			m.Log.V(1).Info(fmt.Sprintf("failed to update owner reference for %s error: %s", val, err.Error()))
+		}
+	}
+
+	marshaledPod, err := m.InjectSidecar(pod, val)
```
AlexsJones, line 77: We either should make all the object creation a function return or none of it, to stick to a single clear way of writing

AlexsJones (changes requested):

AlexsJones (approved):

webhooks/mutating_admission_webhook.go
thschue: As discussed, we will stick with this at the moment

PR #23, opened 2022-05-31
feat: clean up configmaps

### LINKED ISSUE #4: [Feature] [Architecture] [Question] Finalization and deprovisioning
When deleting a custom resource definition we need to decide whether that will delete the mounted configmaps or not.
It would have an impact on running workloads that could be mitigated by flagd caching in the last good spec in running memory.

Thoughts welcome.

thschue:
closes #4


controllers/featureflagconfiguration_controller.go
```diff
@@ -162,3 +179,32 @@ func (r *FeatureFlagConfigurationReconciler) finishReconcile(err error, requeueI
...
+func (r *FeatureFlagConfigurationReconciler) CheckOwnerReference(ff *configv1alpha1.FeatureFlagConfiguration, cm corev1.ConfigMap) bool {
+	for _, cmOwner := range cm.OwnerReferences {
+		if cmOwner.UID == r.GetFfReference(ff).UID {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *FeatureFlagConfigurationReconciler) falseVal() *bool {
```
AlexsJones, line 202: Can you please make a `pkg/utils/utils.go` for these repetitive boolean pointers?

webhooks/mutating_admission_webhook.go
```diff
@@ -94,7 +94,7 @@ func (m *PodMutator) InjectDecoder(d *admission.Decoder) error {
 func CheckOwnerReference(pod *corev1.Pod, cm corev1.ConfigMap) bool {
```
AlexsJones, line 94: Does this need to be public?

controllers/featureflagconfiguration_controller.go
```diff
@@ -116,8 +117,23 @@ func (r *FeatureFlagConfigurationReconciler) Reconcile(ctx context.Context, req
 		}
 	}
 
-	// Update ConfigMaps
 	for _, cm := range ffConfigMapList {
+		// Append OwnerReference if not set
+		if !r.CheckOwnerReference(ffconf, cm) {
```
AlexsJones, line 122:
```suggestion
if !r.featureFlagResourceIsOwner(ffconf,cm) {
```

can we have a method name more in line with what's going on? It's not a generic owner reference check, it's very specific to the feature flag CR>

webhooks/mutating_admission_webhook.go
```diff
@@ -104,34 +104,37 @@ func CheckOwnerReference(pod *corev1.Pod, cm corev1.ConfigMap) bool {
...
 	ffConfig := configv1alpha1.FeatureFlagConfiguration{}
 	if err := m.Client.Get(ctx, client.ObjectKey{Name: name, Namespace: namespace}, &ffConfig); errors.IsNotFound(err) {
-		return configv1alpha1.FeatureFlagConfigurationSpec{}
+		return configv1alpha1.FeatureFlagConfiguration{}
 	}
-	return ffConfig.Spec
+	return ffConfig
 }
 
 func (m *PodMutator) InjectSidecar(pod *corev1.Pod, configMap string) ([]byte, error) {
```
AlexsJones, line 140: like many of the methods on this object, this can be private

AlexsJones (changes requested):

AlexsJones:
Looks great, just some small details I think would really shape it up

controllers/featureflagconfiguration_controller.go
thschue: done

webhooks/mutating_admission_webhook.go
thschue: no, fixed :-)

controllers/featureflagconfiguration_controller.go
thschue: fixed

webhooks/mutating_admission_webhook.go
thschue: done

PR #26, opened 2022-06-01
feat: create configmap if manually deleted

### LINKED ISSUE #24: [Feature] recreate configmaps when manually deleted
Currently, configmaps do not get reconciled when they are deleted/modified. 

## Possible implementation (in controller)
* Get a list of all pods that are using the feature flag
* When a pod uses a configmap that does not exist -> create it (and set correct ownership)

thschue:
closes #24

### DIFF
--- a/webhooks/mutating_admission_webhook.go
+++ b/webhooks/mutating_admission_webhook.go
@@ -114,19 +114,8 @@ func (m *PodMutator) createConfigMap(ctx context.Context, name string, namespace
 		references = append(references, utils.GetFfReference(&ff))
 	}
 
-	cm := corev1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      name,
-			Namespace: namespace,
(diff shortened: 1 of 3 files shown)


webhooks/mutating_admission_webhook.go
```diff
@@ -114,19 +114,8 @@ func (m *PodMutator) createConfigMap(ctx context.Context, name string, namespace
...
-			Annotations: map[string]string{
-				"openfeature.dev/featureflagconfiguration": name,
-			},
-			OwnerReferences: references,
-		},
-		Data: map[string]string{
-			"config.yaml": ff.Spec.FeatureFlagSpec,
-		},
-	}
+	cm := utils.GenerateFfConfigMap(name, namespace, references, ff.Spec)
```
AlexsJones, line 117: I personally don't like this style, but that alone isn't a good reason

AlexsJones (approved):

PR #29, opened 2022-06-10
Add submodule for schemas, embed in utils

toddbaert:
Added a submodule that references [schemas](https://github.com/open-feature/schemas), a make step that copies it into place, and a directive to embed it into the `utils` package.


config/samples/end-to-end/service/openfeature-nest-example.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
```
AlexsJones, line 1: What do you think about the complete example being a single file? 
I know it's nice to keep them isolated but if we called it `config/samples/demo/nest-example.yaml` with everything self contained it's a bit easier to present to others and conceptually is less moving parts.

AlexsJones (changes requested):

config/samples/end-to-end/service/openfeature-nest-example.yml
toddbaert: I think that's a good idea.

My only concern might be that the service definition is using a pretty basic NodePort configuration; it's assuming no more robust ingress is available. If you think that's OK for demo purposes, I'm fine with it...

AlexsJones (approved):

PR #30, opened 2022-06-12
setting up for providers on the CR

AlexsJones:


webhooks/featureflagconfiguration_webhook.go
```diff
@@ -45,6 +47,16 @@ func (m *FeatureFlagConfigurationValidator) Handle(ctx context.Context, req admi
 			return admission.Denied(fmt.Sprintf("FeatureFlagSpec is not valid JSON: %s", err.Error()))
 		}
 	}
+
+	if config.Spec.Provider.Credentials.Name != "" {
```
thschue, line 51: IMHO this should be part of the controller. I'm not sure what happens in a case where the provider and the flag configurations are applied at the same time and the configuration is faster ...

AlexsJones: Okay, we can move that logic there, I am not wed to this - I just thought putting the check in the webhook made sense

toddbaert: @thschue I'm not sure I understand your concern :sweat_smile: 

> I'm not sure what happens in a case where the provider and the flag configurations are applied at the same time

If the CR instance doesn't yet exist, and it features the optional "provider" field, wouldn't we basically always be creating the provider and the flag spec at the "same time" since they are both parts of the same CR? And doesn't this webhook always run before any controller logic to create said resources, as @AlexsJones seems to be saying?

Can you clarify?

toddbaert (approved):

apis/core/v1alpha1/featureflagconfiguration_types.go
```diff
@@ -29,11 +29,18 @@ import (
 type FeatureFlagConfigurationSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
-
-	// Foo is an example field of FeatureFlagConfiguration. Edit featureflagconfiguration_types.go to remove/update
+	// +optional
+	Provider *FeatureFlagProvider `json:"provider"`
+	// FeatureFlagSpec is the json representation of the feature flag
 	FeatureFlagSpec string `json:"featureFlagSpec,omitempty"`
```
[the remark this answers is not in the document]
AlexsJones: We could explore this in the research repo a little more @toddbaert ?

webhooks/featureflagconfiguration_webhook.go
AlexsJones: I think @thschue was specifically referring to the modification of the object here
```
	if ffconf.Spec.Provider == nil {
		r.Log.Info("No provider specified for FeatureFlagConfiguration, using FlagD")
		ffconf.Spec.Provider = &corev1alpha1.FeatureFlagProvider{
			Name: "flagD",
		}
		if err := r.Update(ctx, ffconf); err != nil {
			r.Log.Error(err, "Failed to update FeatureFlagConfiguration Provider")
			return r.finishReconcile(err, false)
		}
	}

```

This however is in the controller and occurs after the webhook. 


The rationale for checking if the secret exists:

The secret will be mounted against the sidecar ( once we have a provider that requires it ) and if we don't check if it exists, it will break the deployments trying to mount a non-existent object.

PR #33, opened 2022-06-13
fixes issue with credentials

AlexsJones:


config/manager/kustomization.yaml
```diff
@@ -1,16 +1,16 @@
...
+  name: manager-config
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
-  - name: controller
-    newName: controller
-    newTag: latest
+- name: controller
+  newName: tibbar/of-operator
+  newTag: v0.0.5.8
```
toddbaert, line 16: This file changes anytime we deploy with a different image, do we need to .gitignore it?

toddbaert (approved):
LGTM, I'll close mine.

config/manager/kustomization.yaml
AlexsJones: good idea

PR #34, opened 2022-06-17
Use full demo

toddbaert:
Updating to use "full" demo app, latest schema.


beeme1mr (approved):
Makes sense, looks good to me!

PR #44, opened 2022-06-27
added the sync provider and reworked, implemented the injection code …

AlexsJones:
…into flagd


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
```
toddbaert, line 16: I think we .gitignored this but perhaps we neglected to remove it from the cache. It still seems to be getting included.

AlexsJones: Will sort it out, good catch!

webhooks/pod_webhook.go
```diff
@@ -149,12 +151,25 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string) ([]byte, e
 			},
 		},
 	})
+
+	commandSequence := []string{
+		"start", "--uri", "/etc/flagd/config.json",
+	}
+	// FlagD is the default provider name externally
+	if featureFlag.Spec.ServiceProvider != nil && featureFlag.Spec.ServiceProvider.Name != "FlagD" {
```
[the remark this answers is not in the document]
AlexsJones: done

webhooks/pod_webhook.go
```diff
@@ -149,12 +151,25 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string) ([]byte, e
...
+		commandSequence = append(commandSequence, "--service-provider")
+		commandSequence = append(commandSequence, "http")
+	}
+	// Adds the sync provider if it is set
+	if featureFlag.Spec.SyncProvider != nil && featureFlag.Spec.SyncProvider.Name != "" {
+		commandSequence = append(commandSequence, "--sync-provider")
+		commandSequence = append(commandSequence, featureFlag.Spec.SyncProvider.Name)
```
toddbaert, line 166: This may be completely paranoid, but do we need to worry about command injection at all here? Since we're appending input to a command?

toddbaert: Thinking about it more, this is validated by the enum you specified above, so this shouldn't be a problem.

toddbaert (approved):

PR #60, opened 2022-07-17
Updated the schema

AlexsJones:
https://github.com/open-feature/open-feature-operator/pull/60


thomaspoignant (approved):

toddbaert (approved):

toddbaert:
Looks like we added a commit signing requirement? Was that done on purpose?

AlexsJones:
> Looks like we added a commit signing requirement? Was that done on purpose?

I think that was done at the org level

PR #75, opened 2022-08-23
adds nil pointer check

AlexsJones:
Signed-off-by: Alex Jones <alexsimonjones@gmail.com>

### DIFF
--- a/webhooks/pod_webhook.go
+++ b/webhooks/pod_webhook.go
@@ -173,6 +173,11 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
 		FlagDTag = os.Getenv("FLAGD_VERSION")
 	}
 
+	var envs []corev1.EnvVar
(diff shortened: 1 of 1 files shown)


webhooks/pod_webhook.go
```diff
@@ -173,6 +173,11 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
 		FlagDTag = os.Getenv("FLAGD_VERSION")
 	}
 
+	var envs []corev1.EnvVar
+	if featureFlag.Spec.FlagDSpec.Envs != nil {
```
skyerus, line 177:
```suggestion
	if featureFlag.Spec.FlagDSpec != nil {
```
it's the FlagDSpec that's causing the panic

skyerus (changes requested):

webhooks/pod_webhook.go
AlexsJones: ta

skyerus (approved):

PR #77, opened 2022-08-23
Injects configured environment variables into all containers running in the pod

AlexsJones:


AlexsJones:
@james-milligan some how `pkg/utils/flagd-definitions.json` got committed, this is safe to remove right?

james-milligan:
It looks like its still being imported from `"github.com/open-feature/schemas/json"` so shouldn't be any problems

toddbaert (approved):
I haven't tested this, but it seems right to me.

thschue (approved):
LGTM, but I think we should split up the injectSidecar function in the future as it gets a bit too large, in my opinion

webhooks/pod_webhook.go
```diff
@@ -178,6 +178,10 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
 		envs = featureFlag.Spec.FlagDSpec.Envs
 	}
 
+	// Add additional environment variables to existing containers
+	for _, container := range pod.Spec.Containers {
```
skyerus, line 182: `range` copies the slice value, this doesn't change the value in the slice itself.

Something like this would work:
```
for i := 0; i < len(pod.Spec.Containers); i++ {
		cntr := pod.Spec.Containers[i]
		cntr.Env = append(cntr.Env, featureFlag.Spec.FlagDSpec.Envs...)
		pod.Spec.Containers[i] = cntr
	}
```

AlexsJones: You are absolutely right

webhooks/pod_webhook.go
```diff
@@ -178,6 +178,10 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
 		envs = featureFlag.Spec.FlagDSpec.Envs
 	}
 
+	// Add additional environment variables to existing containers
```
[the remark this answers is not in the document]
AlexsJones: Will do this as another PR

webhooks/pod_webhook.go
skyerus: I'm confused, why did this get merged?

AlexsJones: Looks like an error. 
I have the change on my HEAD locally here. Apologies will open and resolve
Screenshot 2022-08-24 at 10 41 22
.

AlexsJones: https://github.com/open-feature/open-feature-operator/pull/79

PR #90, opened 2022-09-21
fix: upgrade dependencies with vulnerabilities

skyerus:
Signed-off-by: Skye Gill <gill.skye95@gmail.com>


AlexsJones (approved):
You're the hero we don't deserve

PR #93, opened 2022-09-26
Revert "fix: upgrade dependencies with vulnerabilities (#90)"

beeme1mr:
This reverts commit 58cdd4ee7c6989e44258bad3e9ed75a3bb465cae.

The upgraded dependencies no longer work with go 1.17. This PR reverts the upgrade until a long term solution is found.


james-milligan:
Ive tested this and 
```
github.com/go-logr/logr v1.2.3
```
does not cause a break, so this dependency can still be updated

AlexsJones (approved):

AlexsJones (commented):
Is this ready for review then?

skyerus:
> Ive tested this and
> 
> ```
> github.com/go-logr/logr v1.2.3
> ```
> 
> does not cause a break, so this dependency can still be updated

Imo a separate PR for this would be better once this has been reverted

PR #121, opened 2022-10-18
fix: build and push to docker registry on tag creation

skyerus:
## This PR

Builds and pushes to docker registry on tag creation.
Previous workflow wasn't being triggered by release-please.


.github/workflows/release-please.yml
```diff
@@ -48,3 +53,46 @@ jobs:
               config/webhook/certificate.yaml
               config/rendered/release.yaml
               config/samples/end-to-end.yaml
+
+  build-oci:
+    needs: release-please
+    runs-on: ubuntu-latest
+    if: ${{ needs.release-please.outputs.release_created }}
```
skyerus, line 60: Perhaps this check isn't needed if we want to publish a new image for every push to main, not only on tag creation @AlexsJones

AlexsJones: nah we dont need too

AlexsJones (approved):

PR #126, opened 2022-10-19
feat: stop creation and mounting of flagd-config config map in case of kubernetes sync-provider

### LINKED ISSUE #117: Stop the creation and mounting of the flagd-config config map in the case of kubernetes sync-provider
Previously `flagd` had no way of ingesting the `featureflagconfiguration` CRD directly so OFO created a config map and mounted it on the `flagd` sidecar.
Now `flagd` can use the `kubernetes` sync-provider, it's able to watch the `featureflagconfiguration` CRD directly. When the sync-provider is set as `kubernetes` the config map shouldn't be created and mounted to `flagd`.

skyerus:
## This PR

Stop the creation and mounting of flagd-config ConfigMap when kubernetes is used as the sync-provider.

### Related Issues

Fixes #117


config/samples/end-to-end.yaml
```diff
@@ -70,7 +70,7 @@ spec:
     spec:
       containers:
         - name: open-feature-demo
-          image: ghcr.io/open-feature/open-feature-demo:latest
+          image: ghcr.io/open-feature/playground-app:v0.1.1
```
skyerus, line 73: The old demo app is too out of date to work with the latest flagd

AlexsJones:
This made me laugh because I was just about to do this work - nice one

config/samples/end-to-end.yaml
AlexsJones: Did the repo change name also?

skyerus: It looks that way, this is where I found the reference: https://github.com/open-feature/playground/blob/main/docker-compose.yaml#L4
Perhaps @beeme1mr can clarify further

AlexsJones (approved):

config/samples/end-to-end.yaml
beeme1mr: Yes, this is correct. I think `ghcr.io/open-feature/open-feature-demo:latest` was manually published and is now out of date. With the recent changes to the playground, the demo container is published as part of the CI process.

PR #129, opened 2022-10-19
docs: simplified operator installation

james-milligan:
Signed-off-by: jamesmilligan <james@omnant.co.uk>

added wait command to readme install instructions


AlexsJones:
I don't think this is necessary

james-milligan:
> I don't think this is necessary

No problem, happy to close

beeme1mr:
What about a wait command for the cert manager? I think that's the issue I've experience in the past.

By the way, this is not a feature. Please use the prefix "docs" in the PR.

james-milligan:
We could provide a code block for the installation of the cert manager with a wait? 

```
kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.9.1/cert-manager.yaml
kubectl wait --for=condition=Available=True deploy --all -n 'cert-manager'
```

toddbaert:
I think that's a good idea. I've seen people run into that issue as well.

AlexsJones:
This feels orthogonal to the one-liner install.

AlexsJones:
> We could provide a code block for the installation of the cert manager with a wait?
> 
> ```
> kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.9.1/cert-manager.yaml
> kubectl wait --for=condition=Available=True deploy --all -n 'cert-manager'
> ```

To play devils advocate: there is an assumption its already installed though, where do we draw the line?

james-milligan:
> To play devils advocate: there is an assumption its already installed though, where do we draw the line?

I would argue that because we are seeing issues arise with first time users that we may need to update the README to put the bumpers up a little, we could just place a code block under the prerequisite as a 'safe' install. If we think its looking a little too bloated we could always move it into a `/docs` and provide a 'first time installation' link at the top of the readme. We can push the line out a little further this way without reducing the value of the remainder for the README.

Personally i think the instructions should 'just work' regardless of how quickly they have been copy and pasted, which is the reasoning for the initial line change and for the updates to allow for the one liner. The one liner works well but only for those who have already cloned the repo.

AlexsJones:
> > To play devils advocate: there is an assumption its already installed though, where do we draw the line?
> 
> I would argue that because we are seeing issues arise with first time users that we may need to update the README to put the bumpers up a little, we could just place a code block under the prerequisite as a 'safe' install. If we think its looking a little too bloated we could always move it into a `/docs` and provide a 'first time installation' link at the top of the readme. We can push the line out a little further this way without reducing the value of the remainder for the README.
> 
> Personally i think the instructions should 'just work' regardless of how quickly they have been copy and pasted, which is the reasoning for the initial line change and for the updates to allow for the one liner. The one liner works well but only for those who have already cloned the repo.

https://k3s.io/

If you look at the get k3s command, it says in a comment `#wait 30 seconds` why can't we do something like that? 

Alternatively let's right a shell script that is an installer

e.g. `curl -sfL https://get.open-feature-operator.io | sh -`

james-milligan:
I think a shell script would be a great idea

hack/install-operator.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.9.1/cert-manager.yaml
+kubectl wait --for=condition=Available=True deploy --all -n 'cert-manager'
+kubectl create ns 'open-feature-operator-system' --dry-run=client -o yaml | kubectl apply -f -
+kubectl apply -f https://github.com/open-feature/open-feature-operator/releases/download/v0.2.5/certificate.yaml
```
AlexsJones, line 5: Doesnt this need release please annotations?

james-milligan:
closing in favour of  #206

PR #134, opened 2022-10-20
fix: CVE-2022-32149

toddbaert:
Will create a proper release for the mentioned fix.


AlexsJones (approved):

toddbaert:
The fix commits weren't properly semantic (fix: xxx). This means no release was prompted, because they were not detected as "user-facing" commits. The manual update of the manefest then really confused release please, because it couldn't find the reference `0.2.6` release.

PR #196, opened 2022-10-28
chore: logging/end-to-end improvement

AlexsJones:
Adding a bit more logging in the webhook and namespacing the demo

### DIFF
--- a/config/samples/end-to-end.yaml
+++ b/config/samples/end-to-end.yaml
@@ -1,9 +1,14 @@
 # A basic flag custom resource
-
+apiVersion: v1
+kind: Namespace
+metadata:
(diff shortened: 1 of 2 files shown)


justinabrahms (approved):
It's not clear to me what the logging V is for. I assumed version which doesn't make sense why we'd have different values in the same code.

AlexsJones:
> It's not clear to me what the logging V is for. I assumed version which doesn't make sense why we'd have different values in the same code.

It's from the library https://github.com/go-logr/logr to enable verbosity levels - typically this sort of approach is similar to `klog` in the Kubernetes API where 1 would be info level and 8 is trace. I agree though it might feel a bit unintuitive

Issue #199, opened 2022-11-01
bug: certificate configuration mismatch in helm


james-milligan:
Deploy using the README instructions for Helm, then deploy a pod with flagd annotations gives the following error.
```
Error from server (InternalError): error when creating "deployment/flagd-flag-spec.yaml": Internal error occurred: failed calling webhook "validate.featureflagconfiguration.openfeature.dev": failed to call webhook: Post "https://ofo-webhook-service.open-feature-operator-system.svc:443/validate-v1alpha1-featureflagconfiguration?timeout=10s": x509: certificate signed by unknown authority
```
Seems to be an issue with the certificate configuration from helm / kustomize.

Will remove the README instructions for helm usage until this is resolved

Issue #202, opened 2022-11-01
feat: Add configuration for clusterRoleBindingName in webhooks package


james-milligan:
The `clusterRoleBindingName` value should to be made configurable, potentially through an environment variable

Found in `webhooks/pod_webhook.go`

james-milligan:
no longer required to fix helm deployment

PR #203, opened 2022-11-01
fix: Helm configuration fix

james-milligan:
## This PR

- fixes helm configuration
- adds hardcoded rbac prefix to prevent helm breaking the referecne in `webhooks/pod_webhook.go`

### Related Issues

Fixes https://github.com/open-feature/open-feature-operator/issues/199

### Follow-up Tasks
Hardcoded rbac prefix value should be removed in favour of making the `webhooks/pod_webhook.go` value configurable.
https://github.com/open-feature/open-feature-operator/issues/202

### How to test

follow the README instructions for helm deployment, then deploy the end-to-end example, there should be no errors and app should behave correctly.

### DIFF
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -35,7 +35,9 @@ spec:
         - /manager
         env:
         - name: FLAGD_VERSION
-          value: latest
+          value: INPUT_FLAGD_VERSION
+        - name: CLUSTER_ROLE_BINDING_PREFIX
+          value: {{ include "chart.fullname" . }}
         image: {{ .Values.controllerManager.manager.image.repository }}:{{ .Values.controllerManager.manager.image.tag
           | default .Chart.AppVersion }}
         imagePullPolicy: Always
--- a/chart/templates/deployment.yaml
+++ b/chart/templates/deployment.yaml
@@ -35,7 +35,7 @@ spec:
         - /manager
         env:
         - name: FLAGD_VERSION
-          value: INPUT_FLAGD_VERSION
+          value: latest
         image: {{ .Values.controllerManager.manager.image.repository }}:{{ .Values.controllerManager.manager.image.tag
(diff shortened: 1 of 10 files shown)


chart/templates/deployment.yaml
```diff
@@ -35,7 +35,7 @@ spec:
         - /manager
         env:
         - name: FLAGD_VERSION
-          value: INPUT_FLAGD_VERSION
+          value: latest
```
toddbaert, line 38: Is this change intended

There seems to be a shell script for `sed`ing this in `hack/`.

toddbaert (approved):
in `main`, If I install via the helm command included here, at the point I deploy a workload (and a webhook is called) I get:

```
Error from server (InternalError): error when creating "end-to-end.yaml": Internal error occurred: failed calling webhook "validate.featureflagconfiguration.openfeature.dev": failed to call webhook: Post "https://ofo-webhook-service.open-feature-operator-system.svc:443/validate-v1alpha1-featureflagconfiguration?timeout=10s": x509: certificate is valid for ofo-$(SERVICE_NAME).$(SERVICE_NAMESPACE).svc, ofo-$(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local, not ofo-webhook-service.open-feature-operator-system.svc
```
It seems like the transformation issue we had previously remains in the case of helm.

This PR resolves this issue for me. I think eventually we should replace all `kustomize` usage` with `helm` - having both seems a bit wonky - but that's just my opinion, maybe there's a good reason we should keep both.

chart/templates/deployment.yaml
AlexsJones: Please don't do this, that value is set intentionally

AlexsJones (changes requested):
Let's meet to discuss, hard coding this doesn't feel right

james-milligan:
closing in favour of  #206

Issue #201, opened 2022-11-01
bug: end-to-end example described in readme no longer works


skyerus:
The step on [this line](https://github.com/open-feature/open-feature-operator/blob/main/README.md?plain=1#L152) fails:

```
kubectl -n open-feature-operator-system apply -f config/samples/end-to-end.yaml
namespace/open-feature-demo created
the namespace from the provided object "open-feature-demo" does not match the namespace "open-feature-operator-system". You must pass '--namespace=open-feature-demo' to perform this operation.
the namespace from the provided object "open-feature-demo" does not match the namespace "open-feature-operator-system". You must pass '--namespace=open-feature-demo' to perform this operation.
the namespace from the provided object "open-feature-demo" does not match the namespace "open-feature-operator-system". You must pass '--namespace=open-feature-demo' to perform this operation.
the namespace from the provided object "open-feature-demo" does not match the namespace "open-feature-operator-system". You must pass '--namespace=open-feature-demo' to perform this operation.
```

A namespace was [introduced in the end-to-end.yaml recently](https://github.com/open-feature/open-feature-operator/commit/909d8bac4beaef3536e97b911ddf44d2a950a959). Awaiting to hear the motivation behind this change before finding a solution. 

cc @AlexsJones

PR #207, opened 2022-11-02
feat: structured the featureflagconfiguration CRD

### LINKED ISSUE #72: Improve flagD CRD Structure
## Overview

The current flagD CRD expects `featureFlagSpec` as a string. Various tools (i.e. Lens) use the OpenAPI V3 Schema exposed by Kubernetes to provide basic configuration UIs and input validation. It may be possible to update the current CRD to include the flagD flag configuration json schema.

## Resources

- https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/
- https://www.pulumi.com/blog/introducing-crd2pulumi/

https://github.com/open-feature/open-feature-operator/issues/72#tasklist-block-cde58e47-d88d-41f4-b5ca-5a43c6f5f0c0

skyerus:
## This PR

Introduces v1alpha2 version of the FeatureFlagConfiguration custom resource definition.
In this version the FeatureFlagSpec field is structured (aside from `variants` and `targeting`, these fields are unstructured objects due to the structure/types being loose), whereas in v1alpha1 the FeatureFlagSpec is a string of json.

This is backwards compatible and makes use of the conversion webhook to effectively convert and store any v1alpha2 CRDs as v1alpha1.

### Related Issues

Fixes #72


config/samples/end-to-end.yaml
```diff
@@ -49,16 +37,12 @@ spec:
...
-        }
-      }
-    }
 ---
 # Deployment of a demo-app using our custom resource
 apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: open-feature-demo-deployment
-  namespace: open-feature-demo
```
skyerus, line 61: Removed the namespaces for now, related to https://github.com/open-feature/open-feature-operator/issues/201

skyerus:
Converted to draft to ensure it's merged at the same time as the flagd component. Ready for review

toddbaert:
I think this is a big improvement, even if just for the syntax support in editors, but for a lot of other reasons. I will look at this and the referenced PR today.

config/crd/bases/core.openfeature.dev_featureflagconfigurations.yaml
```diff
@@ -39,8 +39,38 @@ spec:
             properties:
               featureFlagSpec:
                 description: FeatureFlagSpec is the json representation of the feature
-                  flag
+                  flag specification
                 type: string
+              featureFlagSpecV2:
```
beeme1mr, line 44: It doesn't look like v2 supports `$evaluators`. Could you please make sure that's supported?

apis/core/v1alpha1/featureflagconfiguration_types.go
```diff
@@ -38,15 +41,56 @@ type FeatureFlagConfigurationSpec struct {
...
 	Envs []corev1.EnvVar `json:"envs"`
 }
 
+type FeatureFlagSpec struct {
+	Flags map[string]FlagSpec `json:"flags"`
+}
+
+type FlagSpec struct {
+	// +kubebuilder:validation:Enum=ENABLED;DISABLED
+	State string `json:"state"`
```
toddbaert, line 80: Semantically we'd probably like this to be an limited set of values (`ENABLED`/`DISABLED`) but I think last I checked it was difficult to unmarshall JSON into enums, so it might not be possible/worth it.

toddbaert: Oh, I see the `kubebuilder` at least validates it, so that seems good enough.

config/crd/bases/core.openfeature.dev_featureflagconfigurations.yaml
toddbaert: I think you're right, good catch: https://github.com/open-feature/flagd/blob/019ab1de02c587c189ca71b68999cb1de44d88f6/config/samples/example_flags.json

toddbaert:
I approve this in spirit, but I think @beeme1mr is correct that you might need to add `$evaluators` to the schema.... Probably just as a schemaless object.

toddbaert (approved):
I'm going to preemptively approve this but I think @beeme1mr is correct that you might need to add `$evaluators` to the schema.... Probably just as a schemaless object.

config/crd/bases/core.openfeature.dev_featureflagconfigurations.yaml
beeme1mr: One last thought on this. Would it be possible to bump the API version from `core.openfeature.dev/v1alpha1` to `core.openfeature.dev/v1alpha2` or `core.openfeature.dev/v1beta1` and use that as an opportunity to update the `featureFlagSpec` property? That may be a clean way to upgrade the CRD while maintaining backwards compatibility.

apis/core/v1alpha1/featureflagconfiguration_types.go
```diff
@@ -38,15 +41,56 @@ type FeatureFlagConfigurationSpec struct {
 	// +optional
 	// +nullable
 	FlagDSpec *FlagDSpec `json:"flagDSpec"`
-	// FeatureFlagSpec is the json representation of the feature flag
-	FeatureFlagSpec string `json:"featureFlagSpec,omitempty"`
+	// FeatureFlagSpec is the json representation of the feature flag specification
+	// Deprecated: use FeatureFlagSpecV2
+	FeatureFlagSpec *string `json:"featureFlagSpec,omitempty"`
+	// FeatureFlagSpec is the structured representation of the feature flag specification
+	FeatureFlagSpecV2 *FeatureFlagSpec `json:"featureFlagSpecV2,omitempty"`
```
AlexsJones, line 48: It makes more sense to move this to v1valpha2 than support two versions of an API in a single resource denition

AlexsJones (changes requested):
The ambition to structure the JSON is good but trying to support two version of an API inside a single resource is chaotic. I advise you rework this to `v1alpha2` it also means we will need either an upgrade path or a second reconcile loop for that API version.

AlexsJones:
I also wanted to add that, given we're in an alpha state in terms of Semver, it probably makes sense to make this a breaking change and to not try and support two versions of the API at all. I'd recommend modifying the CRD objectively and not worry about using an `v1alphav2`

skyerus:
Thanks for the feedback, I agree that my initial approach was unwise. I've been working on incrementing the version of the CRD in the correct manor, it's not far from completion.
Despite being in alpha I figured that it'd be painful to introduce this as a breaking change as it may create friction for people who, for example, update their version of OFO but not flagd (and vice-versa). I'd prefer to keep it as a non-breaking change for this reason.

config/samples/end-to-end.yaml
beeme1mr: Does this need to be addressed before we merge?

beeme1mr:
Please make sure to update examples and the readme with the new CRD version.

config/samples/end-to-end.yaml
skyerus: No, there is a separate issue addressing this

README.md
```diff
@@ -107,25 +107,20 @@ When wishing to leverage feature flagging within the local pod, the following st
 _See [here](config/samples/crds/custom_provider.yaml) for additional custom resource parameters_
 
 ```
-apiVersion: core.openfeature.dev/v1alpha1
+apiVersion: core.openfeature.dev/v1beta1
```
AlexsJones, line 110: Can we not move to `v1alpha2` ? going to beta starts to lower the amount of wiggle room we have, the next API version will be graduated to `v1`

PROJECT
```diff
@@ -14,4 +14,15 @@ resources:
   kind: FeatureFlagConfiguration
   path: github.com/open-feature/open-feature-operator/apis/core/v1alpha1
   version: v1alpha1
+  webhooks:
```
AlexsJones, line 17: This isn't necessary

apis/core/v1beta1/featureflagconfiguration_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// FeatureFlagSpec is the structured representation of the feature flag specification
+	FeatureFlagSpec FeatureFlagSpec `json:"featureFlagSpec,omitempty"`
+}
+
+type FlagDSpec struct {
+	// +optional
+	Envs []corev1.EnvVar `json:"envs"`
+}
+
+type FeatureFlagSpec struct {
```
AlexsJones, line 51: Just as a counterpoint to creating a new resource @skyerus , I think you could have also gone down this route:

```
type FeatureFlagConfigurationSpec struct {


Styling with Markdown is supported
        // +kubebuilder:deprecatedversion:warning="core.openfeature.dev/v1alpha1 FeatureFlagSpec is deprecated"
       // FeatureFlagSpec string `json:"featureFlagSpec, omityempty"`

	Flags map[string]FlagSpec `json:"flags"`
	// +optional
	// +kubebuilder:validation:Schemaless
	// +kubebuilder:pruning:PreserveUnknownFields
	// +kubebuilder:validation:Type=object
	Evaluators json.RawMessage `json:"$evaluators,omitempty"`
}
```

AlexsJones (approved):

apis/core/v1alpha2/featureflagconfiguration_conversion.go
```diff
@@ -0,0 +1,89 @@
...
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(ffc).
+		Complete()
+}
+
+func (src *FeatureFlagConfiguration) ConvertTo(dstRaw conversion.Hub) error {
+	dst := dstRaw.(*v1alpha1.FeatureFlagConfiguration)
+
+	dst.ObjectMeta = src.ObjectMeta
+	if src.Spec.ServiceProvider != nil {
```
toddbaert, line 37: Do we have to check that `.Spec` is not `nil`, or will existing validation handle that?

config/samples/end-to-end.yaml
```diff
@@ -10,48 +10,39 @@ metadata:
...
+          "on": true
+          "off": false
+        defaultVariant: "on"
+      hex-color:
+        state: "ENABLED"
+        variants:
+          "red": "CC0000"
+          "green": "00CC00"
+          "blue": "0000CC"
+          "yellow": "yellow"
```
toddbaert, line 27: nitpick, but I don't think we need quotes on any of these fields, do we?

config/manager/kustomization.yaml
```diff
@@ -8,7 +8,7 @@ configMapGenerator:
 - files:
   - controller_manager_config.yaml
   name: manager-config
-apiVersion: kustomize.config.k8s.io/v1beta1
```
toddbaert, line 11: I think this was accidental, probably the result of a coincidental bulk replace?

toddbaert (changes requested):
https://github.com/open-feature/open-feature-operator/pull/207/files#r1018167259

apis/core/v1alpha2/featureflagconfiguration_conversion.go
skyerus: `.Spec` isn't a pointer type so a nil check isn't needed

config/samples/end-to-end.yaml
skyerus: Correct, I'll remove them.
Interestingly,
```
variants:
          "on": true
          "off": false
``` 
does require quotes otherwise on/off gets parsed as true/false.

toddbaert (approved):

config/samples/end-to-end.yaml
skyerus: Unfortunately, a relic of an archaic version of yaml

PR #212, opened 2022-11-03
fix: added namespace configuration for helm install using -n

james-milligan:
## This PR

- adds a kustomize overlay to allow for helm configuration keys to be injected into the rendered yaml.
- updated helm configuration to restore expected functionality to `-n namespace-name`.
  
This pattern can also be used for number replacements following the example below:
```
containerPort: 0{{ .Values.numberValue }}
```
the `0` is required to pass linting when the strategic merge runs, but is dropped when the yaml is deployed
### How to test

package the helm chart with the following command:
```
KUSTOMIZE_PATH=config/overlays/helm IMG=ghcr.io/open-feature/open-feature-operator:main make helm-package
```
install the helm chart with the following commands (only one deployment may exist at a time)
```
cd chart && helm upgrade ofo . --install -n test
```
kubectl create ns test
expected output: ofo deployed to `test` namespace, will fail if namespace does not already exist 
```
helm upgrade ofo . --install
```
expected output: ofo deployed to the default `open-feature-operator-system` namespace


config/overlays/helm/kustomization.yaml
```diff
@@ -0,0 +1,8 @@
+bases:
+  - ../../default
+
+namespace: |-
+  {{ include "chart.namespace" . }}
+
+patchesStrategicMerge:
```
toddbaert, line 7: I'm having a tough time understanding this.

toddbaert:
> KUSTOMIZE_PATH=config/overlays/helm IMG=ghcr.io/open-feature/open-feature-operator:main make helm-package

@james-milligan  can we add update the `KUSTOMIZE_PATH` var in the Makefile so the make command becomes less unwieldy?

toddbaert (approved):
Tested both scenarios as described on clean clusters and both worked as expected.

See my note here: https://github.com/open-feature/open-feature-operator/pull/212#issuecomment-1302929373

config/overlays/helm/kustomization.yaml
james-milligan: Ive added comments to describe this behaviour

Makefile
```diff
@@ -3,7 +3,7 @@
 IMG ?= controller:latest
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 FLAGD_VERSION=v0.2.5
-CHART_VERSION=v0.2.16# x-release-please-version
+CHART_VERSION=v0.2.16# x-release-please-vers ion
```
AlexsJones, line 6: That looks wrong

Makefile
```diff
@@ -105,7 +105,7 @@ uninstall: manifests kustomize ## Uninstall CRDs from the K8s cluster specified
 release-manifests: manifests kustomize
 	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG}
 	mkdir -p config/rendered/
-	$(KUSTOMIZE) build config/default > config/rendered/release.yaml
+	$(KUSTOMIZE) build config/overlays/helm > config/rendered/release.yaml
```
AlexsJones, line 108: Doesn't adding this here now break the release.yaml inside of the artifacts on the repo?

james-milligan: updated, added if statement from a `KUSTOMIZE_OVERLAY` variable, the helm build writes directly to `chart/templates/rendered.yaml` now

Makefile
james-milligan: resoloved

AlexsJones (approved):

PR #216, opened 2022-11-09
fix: nil pointer dereference

### LINKED ISSUE #215: bug: nil pointer dereference
If flagDSpec isn't provided in the FeatureFlagConfigurationSpec a nil pointer dereference panic occurs.

skyerus:
## This PR

Fixes a nil pointer dereference when flagDSpec wasn't provided.

### Related Issues

Fixes #215

### DIFF
--- a/webhooks/pod_webhook.go
+++ b/webhooks/pod_webhook.go
@@ -45,6 +45,12 @@ type PodMutator struct {
 
 // Handle injects the flagd sidecar (if the prerequisites are all met)
 func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	defer func() {
+		if err := recover(); err != nil {
+			admission.Errored(http.StatusInternalServerError, fmt.Errorf("%v", err))
+		}
+	}()
+
 	pod := &corev1.Pod{}
 	err := m.decoder.Decode(req, pod)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


webhooks/pod_webhook.go
```diff
@@ -45,6 +45,12 @@ type PodMutator struct {
 
 // Handle injects the flagd sidecar (if the prerequisites are all met)
 func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	defer func() {
```
skyerus, line 48: I wanted to create a test to capture this behaviour. Unfortunately without this block the test would pass regardless of whether the nil check was there or not. The code would panic but because the panic wasn't being caught it'd fail silently (in the eyes of the test).
This safeguard might be overly cautious, perhaps including a test for this case is overkill and just including the nil fix would suffice.
I couldn't come to a decision on this, some other opinions would be appreciated @AlexsJones @toddbaert @beeme1mr @james-milligan

skyerus: The panic can't be caught at a higher level (e.g. in the test suite) because `recover()` can't catch panics in goroutines.

AlexsJones: I think it's overkill personally, but I commend your efforts to improve things.

skyerus: I think it probably is too, my main concern is that something like this can slip through the cracks and not raise any alarms with test failures.

PR #220, opened 2022-11-11
chore: run e2e test as a part of the github workflow

### LINKED ISSUE #205: Implement end-to-end test
A kind cluster can be created via [a github action](https://github.com/marketplace/actions/kind-cluster).

Inject a simple lightweight workload that curls flagd to assert OFO is functioning as expected.

skyerus:
## This PR

Includes a simple e2e test that deploys the operator & spins up an nginx reverse proxy as a direct passthrough to the flagd sidecar.
The test curls the reverse proxy to ensure we get a sensible response from flagd.

The e2e test has been appended to the github workflow triggered on pull requests to main.

### Related Issues

Fixes #205


config/manager/manager.yaml
```diff
@@ -31,7 +31,7 @@ spec:
         - /manager
         args:
         - --leader-elect
-        imagePullPolicy: Always
+        imagePullPolicy: IfNotPresent
```
skyerus, line 34: Are there concerns with this change?

Without it the step that loads the local image into the kind cluster is redundant because it just attempts to pull regardless. If this isn't an acceptable change another possibility could be to push the built image to a registry (as discussed and intentionally avoided previously).

@toddbaert @beeme1mr @AlexsJones @james-milligan

.github/workflows/pr-checks.yml
```diff
@@ -7,7 +7,7 @@ on:
   pull_request:
```
beeme1mr, line 7: By the way, it should be easy enough to avoid running these tests if only markdown files are changed.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths

skyerus: Thanks, done

config/manager/manager.yaml
toddbaert: I think in any non-dev scenario, the images would have a real version appended, so I think this is fine. If anything it could impact local dev a bit, but with the safeguards this PR adds I think that's worth it.

test/e2e/e2e.yml
```diff
@@ -0,0 +1,93 @@
+# A basic flag custom resource
```
toddbaert, line 1: I think it may be worth adding a "top level" comment explaining what this is (testing the mutating injection webhook by proxying through an nginx, which is also the "injected" workload).

beeme1mr (approved):
This looks great! Thanks for putting this together.

A future enhancement to this E2E test could be to make sure changes made to the CR are reflected in FlagD.

test/e2e/e2e.yml
skyerus: Good shout, done

skyerus:
Thanks for the feedback @toddbaert @beeme1mr, I've created an issue as a follow up for the CR reconciliation: https://github.com/open-feature/open-feature-operator/issues/236

Issue #219, opened 2022-11-10
[Bug] Helm releases are not persisting during creation as part of the release process


AlexsJones:
We need to save the chart created as part of the release process in `charts`

beeme1mr:
I think we may want to consider using the [Helm Release Action](https://helm.sh/docs/howto/chart_releaser_action/) here. Basically, the charts will be published to the `gh-pages` branch and we would just expose that branch as our GitHub page.

This would have the following benefits:

- Avoid branch protection issues that are difficult to workaround.
- Remove charts from the main branch
- Follow a documented Helm deployment pattern

AlexsJones:
Would it being on that branch still make the artefacts available?

beeme1mr:
From what I can tell, yes. As long as the GitHub page contains the artifact, everything should work.

AlexsJones:
Is this still open?

beeme1mr:
Yes, it's still an issue. @skyerus was going to look into this tomorrow.

skyerus:
I don't think [Helm release action](https://helm.sh/docs/howto/chart_releaser_action/) is exactly what we're looking for here as it creates a new release which we already have a workflow for (release-please). However, we can hijack the rest of the process and reproduce it in the existing release workflow.

- [ ] Create a branch called `gh-pages` from `main`
- [ ] Change the Github Pages branch setting to `gh-pages` [here](https://github.com/open-feature/open-feature-operator/settings/pages)
- [ ] Create a PR removing the existing published charts from `main`
- [ ] Create a PR merging `main` to `gh-pages` and re-adding the existing published charts
- [ ] Create a PR to `main` appending `release charts` to the existing workflow

PR #233, opened 2022-11-17
chore: remove published charts

skyerus:
## This PR

Removes published charts (to be served from the `gh-pages` branch).

### Related Issues

#219


AlexsJones (approved):
As long we are 100% sure the gh-pages branch has been deployed with the right packages!

skyerus:
> As long we are 100% sure the gh-pages branch has been deployed with the right packages!

We'll change the pages setting to point to `gh-pages` and verify things work before merging this.

beeme1mr (approved):

PR #238, opened 2022-11-18
chore: re-adding charts

skyerus:
## This PR

Brings up to date with main and re-adds charts


beeme1mr (approved):

.github/workflows/release-please.yml
```diff
@@ -132,3 +133,44 @@ jobs:
...
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Merge release to gh-pages
+        run: git merge ${{ needs.release-please.outputs.release_tag_name }}
+
+      - name: Generate helm charts
+        run: make helm-package
```
james-milligan, line 164: Do we need to define the image for the controller here? 
The makefile commands use the IMG argument which defaults to `controller:latest`
I may have missed something here though

PR #240, opened 2022-11-18
fix: include release tag in helm charts publishing

skyerus:
## This PR

Include release tag in helm charts publishing.


beeme1mr:
This seems more like a fix than a chore.

skyerus:
> This seems more like a fix than a chore.

Amended.

james-milligan (approved):

Issue #225, opened 2022-11-14
Set security context on the flagD container


beeme1mr:
## Overview

Kubernetes supports the ability to defined [security context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) at the pod or container level. This can be used to defines privilege and access control settings. 

## Proposal

The OpenFeature Operator (OFO) should set security context at the container level when injecting the sidecar process. The configuration should follow the least privilege principle and justification should be given for the permissions it is granted.

## Requirements

- Identify the minimum permissions required to run flagD
- Apply the security context to the flagD container
- Document justification for the granted permissions

PR #245, opened 2022-11-21
feat: introduced base security context

james-milligan:

## This PR

- adds a base security context to the container definition for flagd:
```
    container name: flagd
    image: ghcr.io/open-feature/flagd:v0.2.5                                      
        allowPrivilegeEscalation: false   
        capabilities: map[drop:[all]]                           
        privileged: false                               
        readOnlyRootFilesystem: true       
        runAsGroup: 65532                               
        runAsNonRoot: true                           
        runAsUser: 65532 
```

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/225

### Follow-up Tasks
It is likely going to worthwhile to allow for the specification of a sec comp profile path, however discussion is required as to how this is going to be handled. One option is for the path for the specified seccomp profile could be passed into configuration via helm.


webhooks/pod_webhook.go
```diff
@@ -287,6 +287,33 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
...
 	})
 	return json.Marshal(pod)
 }
+
+func setSecurityContext() *corev1.SecurityContext {
+	user := int64(65532)
+	group := int64(65532)
+	return &corev1.SecurityContext{
+		// flagd does not require any additional capabilities, no bits set
+		Capabilities: &corev1.Capabilities{ // test the drop all
```
Kavindu-Dodan, line 300: Should we keep the test comment (`// test the drop all`) or was it for testing?

webhooks/pod_webhook.go
```diff
@@ -287,6 +287,33 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
...
+		Privileged: utils.FalseVal(),
+		// Prevents misconfiguration from allowing access to resources on host
+		RunAsNonRoot: utils.TrueVal(),
+		// Prevent container gaining more privileges than its parent process
+		AllowPrivilegeEscalation: utils.FalseVal(),
+		ReadOnlyRootFilesystem:   utils.TrueVal(),
+		// SeccompProfile defines the systems calls that can be made by the container
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: "RuntimeDefault",
+		},
```
toddbaert, line 317: These all make sense to me.

toddbaert (approved):

webhooks/pod_webhook.go
```diff
@@ -287,6 +287,33 @@ func (m *PodMutator) injectSidecar(pod *corev1.Pod, configMap string, featureFla
 				ContainerPort: flagdMetricsPort,
 			},
 		},
+		SecurityContext: setSecurityContext(),
 	})
 	return json.Marshal(pod)
 }
+
+func setSecurityContext() *corev1.SecurityContext {
+	user := int64(65532)
```
skyerus, line 296: comment about why this number would be good

skyerus (approved):

Kavindu-Dodan (approved):

AlexsJones:
@james-milligan have you deployed and tested this? If we're blocking `open` on root directories that would mean we can't use `/etc` for example to mount the config map option ( Remember its still possible to use configmaps and not just the API for K8s )

james-milligan:
>@james-milligan have you deployed and tested this? If we're blocking open on root directories that would mean we can't use /etc for example to mount the config map option ( Remember its still possible to use configmaps and not just the API for K8s )

Ive tested this using this syncProvider config and have found no issues, its my understanding that this flag only prevents the writing to these root directories by the container, although I may have additional configuration issues that is allowing the code to continue behaving normally:
```
apiVersion: core.openfeature.dev/v1alpha1
metadata:
  name: test-config
spec:
  syncProvider: {
    name: filepath
  }
  featureFlagSpec: |
    {
      "flags": {
        "myFlag": {
          "state": "ENABLED",
          "variants": {
            "on": true,
            "off": false
          },
          "defaultVariant": "off"
        }
      }
    }
```
Using a base config kind cluster and the current flagd release

AlexsJones (approved):

PR #253, opened 2022-11-23
chore: added crd configuration reconciliation e2e test

### LINKED ISSUE #236: Create e2e test verifying CR reconciliation
Write a test verifying the flow of changing a flag value in a CRD, applying it and asserting that the flag update is exposed by flagd.

skyerus:
## This PR

Applies changes to CRD and asserts that the changes are reflected in a curl to the nginx reverse proxy to flagd.

### Related Issues

Fixes #236

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ test: manifests generate fmt vet envtest ## Run tests.
 e2e-test: manifests generate fmt vet
 	kubectl -n open-feature-operator-system apply -f ./test/e2e/e2e.yml
 	kubectl wait --for=condition=Available=True deploy --all -n 'open-feature-operator-system'
-	./test/e2e/e2e.sh '{"value":true,"reason":"DEFAULT","variant":"on"}'
+	./test/e2e/run.sh
 
 .PHONY: lint
 lint:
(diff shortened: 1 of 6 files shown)


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: ghcr.io/open-feature/open-feature-operator
```
beeme1mr, line 15: Is this a regression? I thought we needed `newName` to be controller.

config/manager/manager.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         name: manager
         env:
         - name: FLAGD_VERSION
-          value: INPUT_FLAGD_VERSION
+          value: v0.2.5
```
beeme1mr, line 39: Why was this changed made?

config/manager/kustomization.yaml
skyerus: Accidental, removed

config/manager/manager.yaml
skyerus: Accidental, removed

beeme1mr (approved):

james-milligan (approved):

Issue #251, opened 2022-11-22
Support Multiple Kubernetes Syncs


beeme1mr:
## Overview

The OpenFeature Operator (OFO) supports Custom Resources (CR) for FlagD flag definitions. In some situations, it may be beneficial to split the flag definitions between multiple CRs. This can be used to overcome limitations in the max default size of a CR in Kubernetes or to ease configuration management.

## Proposal

Allow a user to define multiple `openfeature.dev/featureflagconfiguration` annotations that are applied to FlagD as arguments. These arguments are applied in the order they were defined, which determines the merge strategy in case of a collision.

## Requirements

- [x] Update the OFO to support multiple openfeature.dev/featureflagconfiguration` annotations
- [x] open-feature/flagd#217

beeme1mr:
Resolve: https://github.com/open-feature/open-feature-operator/issues/17

PR #256, opened 2022-11-23
feat!: Support Multiple Syncs

james-milligan:

corresponding flagd changes [here](https://github.com/open-feature/flagd/pull/222)

## This PR

- updates the `openfeature.dev/featureflagconfiguration` to allow for multiple syncs to be provided to the flagd sidecar, the annotation value is now a comma separated list with each value providing both the namespace and the CRD name separated with a `/`, if no namespace is provided it is assumed that the crd is in the same namespace as the deployment e.g. `"end-to-end, test/end-to-end-2"`
- introduces remote sync implementation

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/251

### How to test
- This is a breaking change and will not work with the existing flagd version, a build from [this PR](https://github.com/open-feature/flagd/pull/222) will be required to test


apis/core/v1alpha1/featureflagconfiguration_types.go
```diff
@@ -51,10 +53,29 @@ type FlagDSpec struct {
 
 type FeatureFlagSyncProvider struct {
 	Name string `json:"name"`
+	// +optional
+	// +nullable
+	RemoteSyncConfiguration *RemoteSyncConfiguration `json:"remoteSyncConfiguration"`
```
skyerus, line 58: We now support multiple versions of the crd (see the v1alpha2 folder). I suppose these will be breaking changes? If yes we would need to introduce a new version of the crd

config/crd/bases/core.openfeature.dev_featureflagconfigurations.yaml
```diff
@@ -240,6 +240,19 @@ spec:
                 properties:
                   name:
                     type: string
+                  remoteSyncConfiguration:
```
toddbaert, line 243: Is http better than "remote"?  "Remote" seems a bit nonspecific.

toddbaert:
is there a reason to go with a comma delimited string over a proper yaml array? Is it just for more backward compatibility? If so, maybe there's another way to achieve it?

webhooks/pod_webhook.go
```diff
@@ -206,63 +224,78 @@ func (m *PodMutator) getFeatureFlag(ctx context.Context, name string, namespace
...
-	}
+	for _, featureFlag := range featureFlags {
+		if featureFlag.Spec.FlagDSpec != nil {
+			if featureFlag.Spec.FlagDSpec.MetricsPort != 0 {
+				flagdMetricsPort = featureFlag.Spec.FlagDSpec.MetricsPort
+			}
+			envs = append(envs, featureFlag.Spec.FlagDSpec.Envs...)
+		}
+		// if remote is explicitly set
+		if featureFlag.Spec.SyncProvider != nil && featureFlag.Spec.SyncProvider.IsRemote() {
```
toddbaert, line 243: it may be simpler to check `featureFlag.Spec.SyncProvider != nil` just once and use nested `if`s, but I'll leave that up to you.

toddbaert:
I really like the `--uri` arguments, they may the flagd command line much more sensible IMHO, but I think we should proceed really carefully with the CRD spec. I'd like to try our best to maintain compatibility but also adhere to conventions (like yaml arrays) and make things easy to understand.

james-milligan:
> is there a reason to go with a comma delimited string over a proper yaml array? Is it just for more backward compatibility? If so, maybe there's another way to achieve it?
> 
> 
> 
> Maybe we could add `openfeature.dev/featureflagconfigurations` (note the "s") which could be an array?
> 
> 
> 
> I think this would be backward compatible... is this a completely dumb suggestion? @james-milligan @skyerus @AlexsJones 

Unfortunately the annotation values must be of type string - I agree an array would be far nicer!

toddbaert:
> > is there a reason to go with a comma delimited string over a proper yaml array? Is it just for more backward compatibility? If so, maybe there's another way to achieve it?
> > Maybe we could add `openfeature.dev/featureflagconfigurations` (note the "s") which could be an array?
> > I think this would be backward compatible... is this a completely dumb suggestion? @james-milligan @skyerus @AlexsJones
> 
> Unfortunately the annotation values must be of type string - I agree an array would be far nicer!

welp, that explains that. :sweat_smile:

test/e2e/e2e.yml
```diff
@@ -59,7 +59,7 @@ spec:
         app: open-feature-e2e-test
       annotations:
         openfeature.dev: "enabled"
-        openfeature.dev/featureflagconfiguration: "end-to-end-test"
+        openfeature.dev/featureflagconfiguration: "open-feature-operator-system.end-to-end-test"
```
AlexsJones, line 62: Can we use `/` as is standard in k8s? e.g. `namespace/service`

config/crd/bases/core.openfeature.dev_featureflagconfigurations.yaml
james-milligan: will update this to `http`

james-milligan:
> Maybe we could add openfeature.dev/featureflagconfigurations (note the "s") which could be an array?

ill update the parsing to assume that the current requests namespace is to be used if none is defined, this will keep it backwards compatible (excluding the CRD changes required to enable the remote/http sync)

webhooks/pod_webhook.go
james-milligan: opted for a switch instead

test/e2e/e2e.yml
james-milligan: ive introduced this change in both PRs

james-milligan:
e2e tests will continue to fail until the corresponding PR in flagd is merged, but I have tested it locally

README.md
```diff
@@ -111,6 +111,7 @@ apiVersion: core.openfeature.dev/v1alpha2
 kind: FeatureFlagConfiguration
 metadata:
   name: featureflagconfiguration-sample
+  namespace: default
```
AlexsJones, line 114: Is it necessary?

webhooks/pod_webhook.go
```diff
@@ -206,63 +224,85 @@ func (m *PodMutator) getFeatureFlag(ctx context.Context, name string, namespace
...
+			)
+			// if http is explicitly set
+		case featureFlag.Spec.SyncProvider.IsHttp():
+			fmt.Printf("FeatureFlagConfiguration %s using http sync implementation\n", featureFlag.Name)
+			if featureFlag.Spec.SyncProvider.HttpSyncConfiguration != nil {
+				commandSequence = append(
+					commandSequence,
+					"--uri",
+					featureFlag.Spec.SyncProvider.HttpSyncConfiguration.Target,
+					"--bearer-token",
```
AlexsJones, line 264: Is bearer-token now mandatory or are we passing a nil value here?

webhooks/pod_webhook.go
```diff
@@ -206,63 +224,85 @@ func (m *PodMutator) getFeatureFlag(ctx context.Context, name string, namespace
...
+			} else {
+				fmt.Printf("FeatureFlagConfiguration %s is missing a httpSyncConfiguration\n", featureFlag.Name)
+			}
+			// if filepath is explicitly set
+		case featureFlag.Spec.SyncProvider.IsFilepath():
+			fmt.Printf("FeatureFlagConfiguration %s using filepath sync implementation\n", featureFlag.Name)
+			commandSequence = append(
+				commandSequence,
+				"--uri",
+				fmt.Sprintf("file://etc/flagd/%s.json", featureFlag.Name),
```
AlexsJones, line 276: Perhaps this should be a constant path?

webhooks/pod_webhook.go
```diff
@@ -206,63 +224,85 @@ func (m *PodMutator) getFeatureFlag(ctx context.Context, name string, namespace
...
-		})
-		volumeMounts = []corev1.VolumeMount{
-			{
-				Name:      "flagd-config",
-				MountPath: "/etc/flagd",
-			},
+			})
+			volumeMounts = append(volumeMounts, corev1.VolumeMount{
+				Name:      featureFlag.Name,
+				MountPath: "/etc/flagd/",
```
AlexsJones, line 290: Again, perhaps this should be constant

webhooks/pod_webhook.go
```diff
@@ -82,40 +84,56 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+	ffConfigs := []*corev1alpha1.FeatureFlagConfiguration{}
+	for _, ffName := range ffNames {
+		ns, name := parseAnnotation(ffName, req.Namespace)
+		if err != nil {
+			m.Log.V(1).Info(fmt.Sprintf("failed to parse annotation %s error: %s", ffName, err.Error()))
+			return admission.Errored(http.StatusBadRequest, err)
 		}
+		// Check to see whether the FeatureFlagConfiguration has service or sync overrides
+		ff := m.getFeatureFlag(ctx, name, ns)
+		if ff.Spec.SyncProvider != nil && !ff.Spec.SyncProvider.IsKubernetes() {
```
AlexsJones, line 96: Here we are saying that if the provider is filepath or http we should create the configmap; does http need it?

README.md
james-milligan: removed

webhooks/pod_webhook.go
james-milligan: I've now wrapped this in an `if`

webhooks/pod_webhook.go
james-milligan: added constant for file path

webhooks/pod_webhook.go
james-milligan: updated this to check for file path sync, rather than non-kubernetes sync

webhooks/pod_webhook.go
```diff
@@ -23,6 +24,7 @@ const (
 	FlagDImagePullPolicy   corev1.PullPolicy = "Always"
 	clusterRoleBindingName string            = "open-feature-operator-flagd-kubernetes-sync"
 	flagdMetricPortEnvVar  string            = "FLAGD_METRICS_PORT"
+	fileSyncMountPath      string            = "etc/flagd/"
```
skyerus, line 27:
```suggestion
	fileSyncMountPath      string            = "/etc/flagd/"
```

AlexsJones (approved):
Nice work young man.
Please make sure this doesn't break the demo/watchman

apis/core/v1alpha2/featureflagconfiguration_conversion.go
```diff
@@ -71,7 +72,15 @@ func (dst *FeatureFlagConfiguration) ConvertFrom(srcRaw conversion.Hub) error {
 	}
 
 	if src.Spec.SyncProvider != nil {
-		dst.Spec.SyncProvider = &FeatureFlagSyncProvider{Name: src.Spec.SyncProvider.Name}
+		dst.Spec.SyncProvider = &FeatureFlagSyncProvider{
```
skyerus, line 75: Do we not need similar logic in the `ConvertTo` func?

skyerus (approved):

PR #259, opened 2022-11-24
refactor(chart): Extract CRD for Helm chart

nlamirault:
## This PR

- CRD directory into the Helm chart (https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)

### Related Issues

Fixes #


skyerus (commented):
What is the context behind this change?
The helm charts are already exposed to artifact hub via https://open-feature.github.io/open-feature-operator/index.yaml (deployed via the `gh-pages` branch)

beeme1mr:
Hi @nlamirault, could you please provide some additional context behind this PR?

nlamirault:
on [Helm documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/), there is 2 methods to manage the CRD : 
- a `crds` directory
- a separate chart

skyerus:
We have used [this documented workflow](https://helm.sh/docs/howto/chart_releaser_action/) to expose the charts. What is our current workflow missing that is covered by these changes?

nlamirault:
@skyerus it's the correct way.

Helm recommands to have a directory to store CRD. Like that : https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack

toddbaert:
Hey @nlamirault, thanks for bringing this up. I think you're right that it would be better to extract the CRDs from the `rednered.yaml`. This way the CRD support in helm would work as expected according to the [documentation](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/) you linked.

I tested this PR out locally a found a few issues. The change breaks the templating. Specifically, the `spec.conversion.webhook.clientConfig.service.name` and the `spec.conversion.webhook.clientConfig.service.namespace` are incorrect. In the case of the namespace, it's not transformed to match the rest of the chart. This value was templated before:

```
spec:
  conversion:
    strategy: Webhook
    webhook:
      clientConfig:
        service:
          name: open-feature-operator-webhook-service
          namespace: '{{ include "chart.namespace" . }}'      // this is now hard-coded in your change
```

The Kustomize substitution is also failing for the `cert-manager.io/inject-ca-from` annotation. These problems cause 404's or TLS failures trying to do the CRD conversion, 

See the manual changes I had to make to fix the problem:

image

There's probably a few ways to fix this, I'll leave which up to you.

Additionally, I think you should remove the CRD from`chart/templates/rendered.yaml`, since after this change it's duplicated there somewhat confusingly.

toddbaert (changes requested):
see https://github.com/open-feature/open-feature-operator/pull/259#issuecomment-1341527799

nlamirault:
Oh, I missed this feature in CRDs.
CRDs cannot be templated, they should be written as the exact manifest you want to send to your cluster.
I don't know how to do that with Helm.

tillknuesting:
@nlamirault Use Go templates and helm generate?

Issue #255, opened 2022-11-23
(Mostly) invalid annotation: `openfeature.dev`


sftim:
The annotation key `openfeature.dev` is not valid. Public annotation keys should follow a convention:
```
rfc1123-dns-domain-name slash *pchar
```
For example: `my.domain.example/some-annotation-key`
A cluster operator can use private annotation keys, which contain no slash. For example: `some-annotation-key`. Software that's published for use across many clusters should not use private label keys or private annotation keys.

To fix this issue, OpenFeature operator should define a new key for enabling the controller's behavior, and deprecate the old `openfeature.dev` key. It's OK to continue to accept the old key indefinitely.

After the old annotation key has been deprecated for a while, and for extra credit, the admission control mechanism can send a [Warning:](https://kubernetes.io/blog/2020/09/03/warnings/) for any Pod, etc, that is still using the deprecated annotation key.

james-milligan:
The `openfeature.dev: enabled` annotation could be refactored to something along the lines of `openfeature.dev/enabled: enabled` (open to suggestions for a better key as I'm not a fan of this one) to match the convention, however, it may be worthwhile removing this annotation all together. Within the web hook this annotation is is used for an early return in the `Handle` method:
```
	val, ok := pod.GetAnnotations()["openfeature.dev"]
	if ok {
		if val != "enabled" {
			m.Log.V(2).Info("openfeature.dev Annotation is not enabled")
			return admission.Allowed("openfeature is disabled")
		}
	}
```
The parsing of the `openfeature.dev/featureflagconfiguration` annotation then follows a very similar pattern, so we can simply just move the `openfeature.dev Annotation is not enabled` log line into the second annotation check:
```
	val, ok = pod.GetAnnotations()["openfeature.dev/featureflagconfiguration"]
	if !ok {
		return admission.Allowed("FeatureFlagConfiguration not found")
	}
	ffNames := strings.Split(val, ", ")
```
We can then provide a warning for the use of the deprecated  `openfeature.dev` annotation

CC: @AlexsJones @sftim @skyerus @beeme1mr @toddbaert

sftim:
If keeping it, I'd go for `openfeature.dev/enabled: "true"`. That does mean quoting the string value in YAML; however, that's what is common for other boolean annotations.

[...]

PR #262, opened 2022-11-29
refactor: annotation refactor

james-milligan:

## This PR

- deprecates the `openfeature.dev` annotation in favour of `openfeature.dev/enabled`
- logs a warning line if the deprecated annotation is used, but preserves its functionality
- adds doc page defining the effect of each annotation, as well as the deprecation

`DEPRECATED: The openfeature.dev annotation has been superseded by the openfeature.dev/enabled annotation. Docs: https://github.com/open-feature/open-feature-operator/docs/annotations`

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/255

### Follow-up Tasks

The playground app should be updated to reflect this change

### DIFF
--- a/test/e2e/tests/001.flagd-response.sh
+++ b/test/e2e/tests/001.flagd-response.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-./"$(dirname "${BASH_SOURCE[0]}")"/../simple-flag-evaluation.sh '{"value":true,"reason":"DEFAULT","variant":"on"}'
+./"$(dirname "${BASH_SOURCE[0]}")"/../simple-flag-evaluation.sh '{"value":true,"reason":"STATIC","variant":"on"}'
--- a/test/e2e/tests/002.crd-configuration-reconciliation.sh
+++ b/test/e2e/tests/002.crd-configuration-reconciliation.sh
@@ -21,7 +21,7 @@ spec:
     }
 EOF
 
-./"$(dirname "${BASH_SOURCE[0]}")"/../simple-flag-evaluation.sh '{"reason":"DEFAULT","variant":"off"}'
+./"$(dirname "${BASH_SOURCE[0]}")"/../simple-flag-evaluation.sh '{"reason":"STATIC","variant":"off"}'
 EXIT_CODE=$?
 
 kubectl -n open-feature-operator-system apply -f ./test/e2e/e2e.yml > /dev/null # reset state quietly


docs/annotations.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+### `openfeature.dev`
+*This annotation is deprecated in favour of the `openfeature.dev/enabled` annotation and should no longer be used.* 
+
+When a value of `"enabled"` is provided, the operator will inject a flagd sidecar into the annotated pods.
+```
+    metadata:
+    annotations:
+        openfeature.dev/enabled: "true"
```
sftim, line 28: Is this example right?

webhooks/pod_webhook.go
```diff
@@ -58,13 +58,26 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
 
 	// Check enablement
+	enabled := false
 	val, ok := pod.GetAnnotations()["openfeature.dev"]
 	if ok {
-		if val != "enabled" {
-			m.Log.V(2).Info("openfeature.dev Annotation is not enabled")
-			return admission.Allowed("openfeature is disabled")
+		m.Log.V(1).Info("DEPRECATED: The openfeature.dev annotation has been superseded by the openfeature.dev/enabled annotation. " +
+			"Docs: https://github.com/open-feature/open-feature-operator/docs/annotations")
```
sftim, line 65: This isn't the right URL.

webhooks/pod_webhook.go
```diff
@@ -58,13 +58,26 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+	val, ok = pod.GetAnnotations()["openfeature.dev/enabled"]
+	if ok {
+		if val == "true" {
+			enabled = true
+		}
+	}
+
+	if !enabled {
+		m.Log.V(2).Info(`openfeature.dev/enabled annotation is not set to "true"`)
+		return admission.Allowed("openfeature is disabled")
```
sftim, line 79:
```suggestion
		return admission.Allowed("OpenFeature is disabled")
```

sftim (commented):
Some feedback. The Golang code looks reasonable but I haven't checked it thoroughly.

webhooks/pod_webhook.go
james-milligan: updated to `https://github.com/open-feature/open-feature-operator/blob/main/docs/annotations`

docs/annotations.md
james-milligan: Thanks for catching this, I've updated the examples

docs/annotations.md
```diff
@@ -0,0 +1,33 @@
...
+When a value of `"true"` is provided, the operator will inject a flagd sidecar into the annotated pods.  
+Example: 
+```
+    metadata:
+    annotations:
+        openfeature.dev/enabled: "true"
+```
+
+### `openfeature.dev/featureflagconfiguration`
+This annotatation defines the FeatureFlagconfiguration CRD that is to be used by the flagd sidecar, only the name of the CRD should be passed, it is expected that the CRD is deployed to the same `namespace` as the pod.
```
beeme1mr, line 15:
```suggestion
This annotation defines the FeatureFlagConfiguration CRD that is to be used by the flagd sidecar, only the name of the CRD should be passed, it is expected that the CRD is deployed to the same `namespace` as the pod.  
```

AlexsJones (approved):

docs/annotations.md
sftim: **Aside**
It actually _specifies the name_ of a FeatureFlagConfiguration custom resource.

The CRD is what makes the FeatureFlagConfiguration API appear in your cluster. A named FeatureFlagConfiguration is a custom resource, not a custom resource definition. CRDs are not namespaced; the APIs that they define might be.

docs/annotations.md
```diff
@@ -12,7 +12,7 @@ Example:
 ```
 
 ### `openfeature.dev/featureflagconfiguration`
-This annotation defines the FeatureFlagConfiguration CRD that is to be used by the flagd sidecar, only the name of the CRD should be passed, it is expected that the CRD is deployed to the same `namespace` as the pod.  
+This annotation specifies the name of the FeatureFlagConfiguration used to configure the injected flagd sidecar, it is expected that the CR is deployed to the same `namespace` as the pod.
```
AlexsJones, line 15: > it is expected that the CR is deployed to the same `namespace` as the pod.

Is that an important thing to mention? FlagD has an extremely permissive role, it can watch CR's throughout the cluster.

AlexsJones (approved):

PR #272, opened 2022-12-08
feat: add ff shortname, commit httpSyncConfiguration

toddbaert:
This adds a shortname for feature flag configurations. I find not having one a bit tedious when doing surgery on a cluster I've messed up.

This PR also commits some generated code around `httpSyncConfiguration`, which for me at least keeps appearing. I think it was never committed and probably should have been?


config/crd/bases/core.openfeature.dev_featureflagconfigurations.yaml
```diff
@@ -515,6 +517,19 @@ spec:
...
+                    nullable: true
+                    properties:
+                      bearerToken:
+                        type: string
+                      target:
+                        description: Target is the target url for flagd to poll
+                        type: string
+                    required:
+                    - target
+                    type: object
```
toddbaert, line 532: This existed before but was never added to the base yaml.

apis/core/v1alpha2/zz_generated.deepcopy.go
```diff
@@ -240,3 +245,18 @@ func (in *FlagSpec) DeepCopy() *FlagSpec {
...
+
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new HttpSyncConfiguration.
+func (in *HttpSyncConfiguration) DeepCopy() *HttpSyncConfiguration {
+	if in == nil {
+		return nil
+	}
+	out := new(HttpSyncConfiguration)
+	in.DeepCopyInto(out)
+	return out
+}
```
toddbaert, line 262: Everything in this file existed before, but until now wasn't committed.

james-milligan (approved):

Issue #274, opened 2022-12-12
Decouple feature flag spec from flagd config


james-milligan:
## Objective

The `FeatureFlagConfiguration` CRD currently contains flagd specific configurations. This can introduce challenges when a configuration only applies to a subset of workloads that have specific configuration requirements. Decoupling the flagd configuration from the feature flag spec allows for greater flexibility when managing OFO.

## Requirements

- [ ] Create a separate CRD for the `FlagSourceConfiguration`
- [ ] Deprecate the `flagDSpec` property from the `FeatureFlagConfiguration` CRD
- [ ] Add new annotation `openfeature.dev/flagsourceconfiguration`
- [ ] Multiple `FlagSourceConfiguration`s can be applied and merged.
- [ ] Add documentation `FlagSourceConfiguration` and how it's used.
- [ ] Remove requirement for [openfeature.dev/featureflagconfiguration](http://openfeature.dev/featureflagconfiguration) annotation to be present for sidecar injection.

## Question

- What do we want to call the new CRD?
- What does the migration progress look like?

james-milligan:
duplicate of https://github.com/open-feature/open-feature-operator/issues/270

AlexsJones:
Just to help clarify this; can we crystalise the use-case here?

What would be an example of say deployment A & B who are both using the same featureflagConifguration but different options in flagD? https/grpc and different ports perhaps ? 

_Something like that feels reasonable to me, but I want to understand the thinking on the following_

If so, how would we anticipate them selecting differentiated flagDconfigurations? Would we then have to annotate each deployment?
```
deploymentA -> flagDConfigA
deploymentB -> flagDConfigB
```
What happens if the config doesn't exist or is updated? 
Are we going to now watch the flagD config from flagD?

If we introduce a new provider e.g. `flagD2` or an updated provider in a cluster with v1 providers, how would they coexist, would there be a new CR for each provider?

james-milligan:
> Just to help clarify this; can we crystalise the use-case here?
> 
> What would be an example of say deployment A & B who are both using the same featureflagConifguration but different options in flagD? https/grpc and different ports perhaps ?
> 
> _Something like that feels reasonable to me, but I want to understand the thinking on the following_
> 
> If so, how would we anticipate them selecting differentiated flagDconfigurations? Would we then have to annotate each deployment?
> 
> ```
> deploymentA -> flagDConfigA
> deploymentB -> flagDConfigB
> ```
> 
[...]

Issue #271, opened 2022-12-08
Add Helm config options


james-milligan:
The existing OFO helm chart customise overlay can be expanded to include environment variables for the operator to overwrite the default values for the following fields:

- sidecar metrics port
- sidecar exposed port
- sidecar evaluator (may be specific to flagd)
- sidecar image 
- sidecar version tag 

These environment variables would overwrite the existing internal defaults that are bing used by the operator, preventing the need for repeated configuration in the `FeatureFlagConfiguration` `FlagdSpec` property  (`8014`,`8013`, `json`...).

kustomize's `patchesStrategicMerge` can be used to append to the existing environment variable array, allowing these values to be overwritten through the `values.yaml` file, or through the cli: 
```
helm upgrade -i openfeature openfeature/open-feature-operator --set sidecar-image=my-custom-flag-source --set evaluator=yaml
```

beeme1mr:
OFO is now fully configurable via Helm

PR #276, opened 2022-12-12
feat: Decouple feature flag spec from flagd config

james-milligan:
## This PR

- deprecates the `flagDConfig` field of the `FeatureflagConfiguration` crd
- introduces a new CRD called `FlagSourceConfiguration` to handle the config
- new annotation `openfeature.dev/flagsourceconfiguration` and associated documentation

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/274


james-milligan:
Ill expand the test suite once everyone is happy with the approach

config/crd/kustomization.yaml
```diff
@@ -3,17 +3,20 @@
...
 resources:
 - bases/core.openfeature.dev_featureflagconfigurations.yaml
+- bases/core.openfeature.dev_flagdconfigurations.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
 patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
 - patches/webhook_in_featureflagconfigurations.yaml
+#- patches/webhook_in_flagdconfigurations.yaml
```
skyerus, line 13: why are these disabled unlike the featureflagconfigurations webhook?

config/manager/manager.yaml
```diff
@@ -36,7 +36,7 @@ spec:
         name: manager
         env:
         - name: FLAGD_VERSION
-          value: INPUT_FLAGD_VERSION
+          value: v0.2.7
```
skyerus, line 39: shouldn't be checked in

skyerus (commented):
generated charts need to be removed (i can't comment on the file)

config/crd/kustomization.yaml
james-milligan: I have enabled them, missed this on my first pass

controllers/flagdconfiguration_controller.go
```diff
@@ -0,0 +1,62 @@
...
+// the FlagdConfiguration object against the actual cluster state, and then
+// perform operations to make the cluster state reflect the state specified by
+// the user.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.13.0/pkg/reconcile
+func (r *FlagdConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	_ = log.FromContext(ctx)
+
+	// FlagdConfigurations do not currently update cluster state, and only provides configuration for the pod_webhook
```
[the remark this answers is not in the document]
skyerus: Should an update to the `FlagdConfiguration` not trigger a change of flagd's configured state?

index.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
skyerus, line 1: this also shouldn't be checked in

controllers/flagdconfiguration_controller.go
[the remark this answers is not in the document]
james-milligan: I gave this some thought (and am open to other opinions), the configurations which can be passed by the `FlagdConfiguration` all require the container to be restarted for changes to take place as they are provided via arguments or picked up from environment variables on startup, as a result of this I opted to keep the changes as simple as possible, with new configurations being picked up on pod (re)deployment.
I would not expect for a configuration change to take immediate effect, especially if additional changes would also be required by the accompanying containers within the pod (for example an exposed port change would need to be reflected in the client)

index.yaml
james-milligan: removed

controllers/flagdconfiguration_controller.go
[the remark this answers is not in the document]
skyerus: I agree with this approach as long as it's not an anti-pattern and we document it well

beeme1mr (commented):
Did you mean to include a `.DT_Store`? Apparently you're on a Mac 😄

james-milligan:
> Did you mean to include a `.DT_Store`? Apparently you're on a Mac 😄

caught using `git add .`

chart/templates/rendered.yaml
```diff
@@ -527,6 +557,149 @@ spec:
...
+          namespace: '{{ include "chart.namespace" . }}'
+          path: /convert
+      conversionReviewVersions:
+      - v1
+  group: core.openfeature.dev
+  names:
+    kind: FlagdConfiguration
+    listKind: FlagdConfigurationList
+    plural: flagdconfigurations
+    singular: flagdconfiguration
```
toddbaert, line 583: Can we add a `shortName`? `fdc`, maybe?

chart/templates/rendered.yaml
```diff
@@ -247,6 +250,19 @@ spec:
...
+                    nullable: true
+                    properties:
+                      bearerToken:
+                        type: string
+                      target:
+                        description: Target is the target url for flagd to poll
+                        type: string
+                    required:
+                    - target
+                    type: object
```
toddbaert, line 265: I think there are changes in this PR (such as this) which are only appearing because you need to rebase on main.

apis/core/v1alpha1/flagdconfiguration_types.go
```diff
@@ -0,0 +1,165 @@
...
+	// Tag to be appended to the flagd image, defaults to 'main'
+	// +optional
+	Tag string `json:"tag"`
+}
+
+func NewFlagdConfigurationSpec() *FlagdConfigurationSpec {
+	var tag = "main"
+	if os.Getenv("FLAGD_VERSION") != "" {
+		tag = os.Getenv("FLAGD_VERSION")
+	}
```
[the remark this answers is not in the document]
beeme1mr: Should `FLAGD_VERSION` be included in the const above?

docs/flagd_configuration.md
```diff
@@ -0,0 +1,63 @@
+# Flagd Configuration
+
+The injected flagd sidecar is configured using the `FlagdConfiguration` CRD, the `openfeature.dev/flagdconfiguration` annotation is used to assign `Pods` with their respective `FlagdConfiguration` CRs. The annotation value is a comma separated list of values following one of 2 patterns: {NAME} or {NAMESPACE}/{NAME}. If no namespace is provided it is assumed that the CR is within the same namespace as the deployed pod, for example:
```
skyerus, line 3:
```suggestion
The injected flagd sidecar is configured using the `FlagdConfiguration` CRD, the `openfeature.dev/flagdconfiguration` annotation is used to assign `Pods` with their respective `FlagdConfiguration` CRs. The annotation value is a comma separated list of values following one of 2 patterns: {NAME} or {NAMESPACE}/{NAME}. If no namespace is provided, it is assumed that the CR is within the same namespace as the deployed pod, for example:
```

docs/flagd_configuration.md
```diff
@@ -0,0 +1,63 @@
...
+```
+    metadata:
+        namespace: test-ns
+        annotations:
+            openfeature.dev/enabled: "true"
+            openfeature.dev/flagdconfiguration:"config-A, test-ns-2/config-B"
+```
+In this example 2 CRs are being used to configure the injected flagd container, `config-A` (which is assumed to be in the namespace `test-ns`) and `config-B` from the `test-ns-2` namespace, with `config-B` taking precedence in the configuration merge.
```
skyerus, line 11:
```suggestion
In this example, 2 CRs are being used to configure the injected flagd container, `config-A` (which is assumed to be in the namespace `test-ns`) and `config-B` from the `test-ns-2` namespace, with `config-B` taking precedence in the configuration merge.
```

docs/flagd_configuration.md
```diff
@@ -0,0 +1,63 @@
...
+| Port   | Defines the port for flagd to listen on, defaults to 8014        | `int32`        |
+| SocketPath   | Defines the unix socket path to listen on        | `string`       |
+| SyncProviderArgs   | String arguments passed to flagd on startup, documentation can be found [here](https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md)        | `array of strings`, key values separated by `=`, e.g `key=value`       |
+| Evaluator   | Sets the flagd evaluator, defaults to `json`        | `string`       |
+| Image   | Allows for the flagd image to be overridden, defaults to `ghcr.io/open-feature/flagd`        | `string`       |
+| Tag   |  Tag to be appended to the flagd image, defaults to `main`        | `string`       |
+
+## Configuration Merging
+
+When multiple `FlagdConfigurations` are passed the configurations are merged, the latest `CR` will take precedence over the first, with any configuration from the deprecated `FlagDSpec` field of the `FeatureFlagConfiguration` CRD taking the lowest priority.
```
skyerus, line 29:
```suggestion
When multiple `FlagdConfigurations` are passed the configurations are merged, the last `CR` will take precedence over the first, with any configuration from the deprecated `FlagDSpec` field of the `FeatureFlagConfiguration` CRD taking the lowest priority. 
```

PROJECT
```diff
@@ -25,4 +25,13 @@ resources:
   kind: FeatureFlagConfiguration
   path: github.com/open-feature/open-feature-operator/apis/core/v1alpha2
   version: v1alpha2
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: openfeature.dev
+  group: core
+  kind: FlagdConfiguration
```
AlexsJones, line 34: My 2c but the custom resource should be `OpenFeatureOperatorConfigruation` or `OFOConfiguration` because `flagD` is the current in-vogue provider for injecting into the workloads. Having a CRD for each provider doesn't seem scalable, in addition to that, I would also like to put our more esoteric configuration for OFO, especially as we consider allowing for flagD upgrades in the OFO configuration.

docs/flagd_configuration.md
```diff
@@ -0,0 +1,63 @@
...
+    tag: latest
+```
+Config-B:
+```
+apiVersion: core.openfeature.dev/v1alpha2
+kind: FlagdConfiguration
+metadata:
+    name: test-configuration
+spec:
+    Port: 8000
```
toddbaert, line 54:
```suggestion
    port: 8000
```

toddbaert: This casing causes validation errors for me (same as https://github.com/open-feature/open-feature-operator/pull/276/files#r1048735805)

toddbaert:
@james-milligan I tested this locally, and besides the casing of `Port` in your samples, things work for me. 

How I tested:

- setup new cluster
- built your changes and ran `helm install ofo .`
- applied (old style) config [here](https://github.com/open-feature/open-feature-operator/releases/download/v0.2.20/release.yaml)

Everything worked as expected.

Then I updated the config mentioned above to use the new CRD, and...

- setup a new clust
- applied the config (worked as expected)
- intentionally changed the port in the `flagDConfiguration` CRD, killed my workload pod, and saw it start up with an expected error because I intentionally chose a new port that would conflict with an already opened one (8014). This was expected and proves the CRD worked in changing flag's settings

PROJECT
toddbaert: I see your point... but we're not necessarily configuring the operator itself with this CRD are we? We are configuring the `source-of-truth` for the flags in the relevant workload(s). 

What about `FlagSource`? shortname `fs` or something like that?

james-milligan: +1 for `ffs`

I agree with Todd here that these configurations apply to the workloads so something along the lines on `SidecarConfiguration` or `FlagSourceConfiguration` would be more suitable

toddbaert (approved):
I'm approving this but I still think [this](https://github.com/open-feature/open-feature-operator/pull/276#discussion_r1048666403) is a good point. If we end up renaming, I'm happy to do a sanity check again.

PROJECT
beeme1mr: Hey @AlexsJones, does this address your concerns?

james-milligan:
Converting to draft until I return from my Christmas break! 🎅

PROJECT
james-milligan: I have updated the name of the CRD to `FlagSourceConfiguration` along with its annotation

skyerus (approved):

docs/annotations.md
```diff
@@ -23,6 +23,22 @@ Example:
...
+
+### `openfeature.dev/flagsourceconfiguration`
+This annotation specifies the names of the FlagSourceConfigurations used to configure the injected flagd sidecar.
+The annotation value is a comma separated list of values following one of 2 patterns: {NAME} or {NAMESPACE}/{NAME}. 
+If no namespace is provided it is assumed that the CR is within the same namespace as the deployed pod.
+If multiple CRs are provided, they are merged with the latest taking precedence, for example, in the scenario below, `config-B` will take priority in the merge, replacing duplicated values that are set in `config-A`.
+
+Example:
+```
+    metadata:
```
beeme1mr, line 35:
```suggestion
  metadata:
```

beeme1mr:
Hey @toddbaert @AlexsJones, any concerns about this PR? It would be great to get this merged so @james-milligan can [add support for these settings in Helm](https://github.com/open-feature/open-feature-operator/issues/271).

toddbaert (approved):
> Hey @toddbaert @AlexsJones, any concerns about this PR? It would be great to get this merged so @james-milligan can https://github.com/open-feature/open-feature-operator/issues/271.

@beeme1mr My only concern was [this](https://github.com/open-feature/open-feature-operator/pull/276#discussion_r1048666403) which is resolved now. Approved.

AlexsJones (approved):
looks fine

Issue #292, opened 2023-01-10
Flagd sync breaks after chart update


beeme1mr:
## Overview

Updating the OFO via helm causes existing flagd subscriptions to stop working. This seems to be related to the service account used by flagd to subscribe to the Kubernetes API.

## Steps to reproduce

1. Deploy OFO using helm version `0.2.22`
2. Deploy a workload that requires flagd to be injected by OFO
3. Test that flag updates are synced
4. Update OFO to version `0.2.23` using helm
5. Attempt to update a flag configure
6. Confirm that flagd did not receive the flag update
7. Recreate the workload
8. Confirm that the new flagd is receiving flag updates

## Logs

<details>
  <summary>Expand this section to see the logs output by flagd.</summary>

```
         ______   __       ________   _______    ______      
        /_____/\ /_/\     /_______/\ /______/\  /_____/\     
        \::::_\/_\:\ \    \::: _  \ \\::::__\/__\:::_ \ \    
         \:\/___/\\:\ \    \::(_)  \ \\:\ /____/\\:\ \ \ \   
          \:::._\/ \:\ \____\:: __  \ \\:\\_  _\/ \:\ \ \ \  
           \:\ \    \:\/___/\\:.\ \  \ \\:\_\ \ \  \:\/.:| | 
            \_\/     \_____\/ \__\/\__\/ \_____\/   \____/_/                                                         

{"level":"warn","ts":1673279286.986885,"caller":"kubernetes/kubernetes_sync.go:57","msg":"Client not initialised","component":"sync","sync":"kubernetes"}
{"level":"info","ts":1673279286.988456,"caller":"kubernetes/kubernetes_sync.go:104","msg":"Starting kubernetes sync notifier for resource dev/end-to-end","component":"sync","sync":"kubernetes"}
{"level":"info","ts":1673279287.0330749,"caller":"service/connect_service.go:109","msg":"metrics listening at 8014","component":"service"}
{"level":"info","ts":1673279287.1486142,"caller":"kubernetes/kubernetes_sync.go:136","msg":"kube sync notifier event: add dev end-to-end","component":"sync","sync":"kubernetes"}
{"level":"info","ts":1673279287.150566,"caller":"kubernetes/kubernetes_sync.go:136","msg":"kube sync notifier event: add dev end-to-end","component":"sync","sync":"kubernetes"}
[...]

PR #295, opened 2023-01-12
feat: role binding backfill

james-milligan:
## This PR

- backfill the flagd-kubernetes-sync role binding on upgrade

### Related Issues

https://github.com/open-feature/open-feature-operator/issues/292

### Follow-up Tasks

This PR expands one of the options for resolving the bug described in the issue above. On startup a grouroutine is started with a 5 second timeout (to allow cache to be ready), and all pods with the `openfeature.dev/enabled` annotation set to `"true"` will have their service accounts added to the `flagd-kubernetes-sync` role binding, restoring its state to that which it was prior to the upgrade.

### How to test

Upgrade helm and verify that the flagd connection is reestablished.


this.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: core.openfeature.dev/v1alpha2
```
skyerus, line 1: should this be included?

config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
-  newTag: latest
+  newName: ghcr.io/open-feature/open-feature-operator
```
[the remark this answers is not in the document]
james-milligan: resolved

help.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
james-milligan: resolved

this.yaml
james-milligan: resolved

webhooks/pod_webhook.go
```diff
@@ -54,6 +54,7 @@ type PodMutator struct {
 // BackfillPermissions recovers the state of the flagd-kubernetes-sync role binding in the event of upgrade
 func (m *PodMutator) BackfillPermissions(ctx context.Context) {
 	for i := 0; i < 5; i++ {
+		fmt.Println("im running")
```
beeme1mr: 😄

james-milligan: you caught that before I removed it :(

webhooks/pod_webhook.go
```diff
@@ -47,6 +51,38 @@ type PodMutator struct {
 	Log                       logr.Logger
 }
 
+// BackfillPermissions recovers the state of the flagd-kubernetes-sync role binding in the event of upgrade
+func (m *PodMutator) BackfillPermissions(ctx context.Context) {
+	for i := 0; i < 5; i++ {
+		if i == 5 {
```
Kavindu-Dodan, line 57: Let's move this out of the for loop. This gives more focus on the operations we attempt to perform

Plus, at this point our flagd deployments are unstable. Should we avoid operator start?

toddbaert:
I could be missing something, but I still seem to have the issue after and upgrade. Here are the steps I took:

- built/pushed my docker image: `toddbaert/ofo:main`
- built charts: `make build && make helm-package`
- edited rendered.taml to use `toddbaert/ofo:main`
- deployed workload and verified everything worked

Then I applied the fix:

- checked out this branch
- edited `Makefile`, updating version to `v0.2.24`
- built/pushed my docker image: `toddbaert/ofo:fix`
- built charts: `make build && make helm-package`
- edited rendered.taml to use `toddbaert/ofo:fix`
- did `helm upgrade ofo .` from chart dir
- changed a flag value on the workload, saw:

```
"end-to-end\" is forbidden: User \"s │
│ ystem:serviceaccount:default:default\" cannot get resource \"featureflagconfigurations\" in API group \
```

toddbaert:
@james-milligan 

As discussion on slack, the root cause of my issue was that I was using the old invalid annotation `openfeature.dev: "enabled"` not `openfeature.dev/enabled: "true"` (which works as expected).

As you noted, we may want to add support for the deprecated annotation too.

webhooks/pod_webhook.go
james-milligan: This is a timeout allowing for the cache to be set up, which is managed in a separate groutine (unfortunately there is no 'ready' channel to hook into so the loop is required) -  it does not take into account the health of the deployments, and will backfill permissions even if the pod is crashing, i.e. 
```
NAME                             READY   STATUS             RESTARTS     AGE
busybox-curl-2-67d494d8c-4ldw8   1/2     CrashLoopBackOff   1 (2s ago)   8s
```
will still have its service account added as a subject to the role binding if its annotations are valid.

Ive cleaned up this loop, moving the timeout error to a drop out once the loop has finished

james-milligan:
> @james-milligan
> 
> As discussed on on slack, the root cause of my issue was that I was using the old invalid annotation `openfeature.dev: "enabled"` not `openfeature.dev/enabled: "true"` (which works as expected).
> 
> As you noted, we may want to add support for the deprecated annotation too.
> 
> But otherwise, this works as expected. Nice job!

Ive added in support for the `core.openfeature.dev: "enabled"` annotation, as well as expanded the test to include it

webhooks/pod_webhook.go
```diff
@@ -383,3 +415,21 @@ func setSecurityContext() *corev1.SecurityContext {
 		},
 	}
 }
+
+func OpenFeatureEnabledAnnotationIndex(o client.Object) []string {
+	val, ok := o.(*corev1.Pod).ObjectMeta.Annotations["openfeature.dev/enabled"]
```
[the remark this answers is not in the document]
james-milligan: you're right, ive resolved this and simplified the remaining statements

webhooks/pod_webhook.go
```diff
@@ -47,6 +51,34 @@ type PodMutator struct {
...
 
+// BackfillPermissions recovers the state of the flagd-kubernetes-sync role binding in the event of upgrade
+func (m *PodMutator) BackfillPermissions(ctx context.Context) {
+	for i := 0; i < 5; i++ {
+		// fetch all pods with the "openfeature.dev/enabled" annotation set to "true"
+		podList := &corev1.PodList{}
+		err := m.Client.List(context.Background(), podList, client.MatchingFields{"metadata.annotations.openfeature.dev/enabled": "true"})
+		if err != nil {
+			if !goErr.Is(err, &cache.ErrCacheNotStarted{}) {
+				m.Log.Error(err, err.Error())
```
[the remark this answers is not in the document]
james-milligan: ive updated this error log to provide more context via the message argument

main.go
```diff
@@ -178,7 +192,18 @@ func main() {
...
-	if err := mgr.Start(ctrl.SetupSignalHandler()); err != nil {
+	ctx := ctrl.SetupSignalHandler()
+	errChan := make(chan error, 1)
+	go func(chan error) {
+		if err := mgr.Start(ctx); err != nil {
+			errChan <- err
+		}
+	}(errChan)
+
+	setupLog.Info("starting backfill")
```
toddbaert, line 203: I think we can improve this message a bit.

james-milligan: updated to `restoring flagd-kubernetes-sync cluster role binding subjects from current cluster state`

toddbaert (approved):

webhooks/pod_webhook.go
```diff
@@ -47,6 +51,43 @@ type PodMutator struct {
 	Log                       logr.Logger
 }
 
+// BackfillPermissions recovers the state of the flagd-kubernetes-sync role binding in the event of upgrade
+func (m *PodMutator) BackfillPermissions(ctx context.Context) {
+	for i := 0; i < 5; i++ {
+		// fetch all pods with the "openfeature.dev/enabled" annotation set to "true"
+		podList := &corev1.PodList{}
+		err := m.Client.List(ctx, podList, client.MatchingFields{"metadata.annotations.openfeature.dev/enabled": "true"})
```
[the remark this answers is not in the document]
james-milligan: introduced constant

james-milligan:
resolved test race condition through introducing a 'complete' channel to the `BackfillPermissions` method, this is not used in `main` as the process can happen asynchronously as the ` mgr.Start(ctx)` method is blocking

Kavindu-Dodan (approved):

skyerus (approved):

PR #303, opened 2023-01-17
test: add retry to async podMutator.BackfillPermissions integration test

james-milligan:
## This PR

- adds retry to integration test for async `podMutator.BackfillPermissions` method


webhooks/run_test.go
```diff
@@ -61,13 +61,17 @@ func run(ctx context.Context, cfg *rest.Config, scheme *runtime.Scheme, opts *en
...
+	go func() {
+		if err := podMutator.BackfillPermissions(ctx); err != nil {
+			errChan <- err
+		}
+	}()
+	go func() {
+		if err := mgr.Start(ctx); err != nil {
+			errChan <- err
+		}
+	}()
```
skyerus, line 74: If both of these end successfully (`err == nil`) then `run()` will never complete

james-milligan: thanks, updated

webhooks/run_test.go
```diff
@@ -61,13 +61,15 @@ func run(ctx context.Context, cfg *rest.Config, scheme *runtime.Scheme, opts *en
...
-	}(ctx, backfillComplete)
+	errChan := make(chan error, 1)
+	go func() {
+		if err := podMutator.BackfillPermissions(ctx); err != nil {
+			errChan <- err
+		}
+	}()
+	go func() {
+		errChan <- err
+	}()
```
skyerus, line 72: This will always return the error on L21 (even if it's nil) and ignore the potential error on L67

james-milligan: resolved, now returns the error from `mgr.Start(ctx)`, if `mgr.Start(ctx)` fails, it will do so before the potential timeout error on L67, so should be safe for returning the appropriate error now. In the case of no error, `mgr.Start(ctx)` will return nil and the function will be unblocked

webhooks/run_test.go
```diff
@@ -61,13 +61,15 @@ func run(ctx context.Context, cfg *rest.Config, scheme *runtime.Scheme, opts *en
...
-	}(ctx, backfillComplete)
+	errChan := make(chan error, 1)
+	go func() {
+		if err := podMutator.BackfillPermissions(ctx); err != nil {
+			errChan <- err
+		}
+	}()
+	go func() {
+		errChan <- mgr.Start(ctx)
+	}()
```
skyerus, line 72: What about the case where mgr.Start returns nil first then the BackfillPermissions returns an error? It'd be an unhandled error, right?

james-milligan: Start won't return until nil the context is closed, the podMutator.BackfillPermissions will always fail before this (or its associated test)

skyerus: Can a comment be added in-code to explain this

webhooks/pod_webhook_test.go
```diff
@@ -164,24 +167,78 @@ func podMutationWebhookCleanup() {
...
-			Kind:      "ServiceAccount",
-			APIGroup:  "",
-			Name:      existingPod1ServiceAccountName,
-			Namespace: mutatePodNamespace,
-		}))
-		Expect(rb.Subjects).To(ContainElement(v1.Subject{
+		// this integration test confirms the proper execution of the  podMutator.BackfillPermissions method
+		// this method is responsible for backfilling the subjects of the open-feature-operator-flagd-kubernetes-sync
+		// cluster role binding, for previously existing pods on startup
+		// a retry is require don this test as the backfilling occurs asynchronously
```
skyerus, line 173:
```suggestion
		// a retry is required on this test as the backfilling occurs asynchronously
```

webhooks/run_test.go
james-milligan: added comment

PR #344, opened 2023-02-07
feat: improve deployment pattern

james-milligan:
## This PR

- deprecates the `openfeature.dev/featureflagconfiguration` annotation, superseding it with `openfeature.dev/flagsourceconfiguration` containing a new configurations + allowing for a many to one relationship to define FeatureFlagconfigurations / configurations
- introduces a new annotation `openfeature.dev/allowkubernetessync: "true"` to label pods requiring permissions to watch the CRs, preventing these permissions being given to all pods
- introduces `v1alpha3` version of `FlagSourceConfiguration` (required to prevent import cycle, this is why the PR is currently a WIP, these versions should be decoupled from each other)
- refactors the pod webhook, is now easier to maintain / read

### Notes

- Requires the appropriate documentation updates
- Requires a few more tests once the versioning is sorted to ensure the conversion functions produce expected behaviour


config/crd/kustomization.yaml
```diff
@@ -11,12 +11,14 @@ patchesStrategicMerge:
 # patches here are for enabling the conversion webhook for each CRD
 - patches/webhook_in_featureflagconfigurations.yaml
 - patches/webhook_in_flagsourceconfigurations.yaml
+#- patches/webhook_in_flagsourceconfigurations.yaml
```
skyerus, line 14: remove

config/crd/kustomization.yaml
```diff
@@ -11,12 +11,14 @@ patchesStrategicMerge:
...
 #+kubebuilder:scaffold:crdkustomizewebhookpatch
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
 - patches/cainjection_in_featureflagconfigurations.yaml
 - patches/cainjection_in_flagsourceconfigurations.yaml
+#- patches/cainjection_in_flagsourceconfigurations.yaml
```
skyerus, line 21: remove

index.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
```
skyerus, line 1: remove file

webhooks/pod_webhook.go
```diff
@@ -149,75 +112,283 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+
+	marshaledPod, err := m.injectSidecar(ctx, pod, flagSourceConfigurationSpec)
+	if err != nil {
+		m.Log.Error(err, "unable to inject flagd sidecar")
+		return admission.Errored(http.StatusInternalServerError, err)
+	}
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+}
+
+func (m *PodMutator) handleFeatureFlagConfigurationAnnotation(ctx context.Context, fcConfig *v1alpha1.FlagSourceConfigurationSpec, ffconfigAnnotation string, defaultNamespace string) error {
```
skyerus, line 154: perhaps this func could be moved to a `pod_webhook_deprecated.go` file?

webhooks/pod_webhook.go
```diff
@@ -304,166 +475,39 @@ func (m *PodMutator) enableClusterRoleBinding(ctx context.Context, pod *corev1.P
...
 
-	cm := corev1alpha1.GenerateFfConfigMap(name, namespace, references, ff.Spec)
+	cm := v1alpha1.GenerateFfConfigMap(name, namespace, references, ff.Spec)
 
 	return m.Client.Create(ctx, &cm)
 }
 
-func (m *PodMutator) getFeatureFlag(ctx context.Context, name string, namespace string) corev1alpha1.FeatureFlagConfiguration {
-	ffConfig := corev1alpha1.FeatureFlagConfiguration{}
+func (m *PodMutator) getFeatureFlag(ctx context.Context, namespace string, name string) v1alpha1.FeatureFlagConfiguration {
```
skyerus, line 495: sanity check that the parameter flipping has been applied everywhere

james-milligan: this is covered by the integration tests

webhooks/pod_webhook.go
james-milligan: great suggestion

toddbaert:
I ran this locally, and I have to say I love it:

- it removes the need for the (not pretty) comma separated annotation value for multiple CR refs
- the CRDs align much better with the logical configuration
- it keeps CRD separation of concerns much cleaner: `FeatureFlagConfiguration` kinds only care about the states of flags, `FlagSourceConfiguration` kinds only care about flagd setup. This also potentially allows for better permissions setup in terms of who can change what types of resource
- the `openfeature.dev/allowkubernetessync` annotation makes a lot of sense

I would advocate we move to this pattern sooner than later for all the above reasons. I just think it's a better model.

### My testing:

I created 2 `FeatureFlagConfiguration`s, and one `FlagSourceConfiguration`, which used a different `syncProvider` for each.

Pretty much everything worked as expected, with a couple small exceptions:

- I couldnt seem to disable the cluster permissions with `openfeature.dev/allowkubernetessync: "false"`. Whether I had previously set this value to `"true"` or not seemed to have no impact. I was expecting that if I set this to `"false"`, I would see some permissions error in the flagd sidecar - I did not - in fact everything worked. Maybe I configured it incorrectly or I'm confused about the purpose of this annotation.
- Expected, I believe, but changes in `FlagSourceConfiguration` kinds had no impact until the workload was cycled - for instance, changing the syncProvider value from `kubernetes` => `filepath` didn't cause flagd to watch the filepath.

### More throughts:

I think that we should _certainly not_ support dynamically changing the mode of flagd operation when a `FlagSourceConfiguration` changes. I think this would add a lot of complexity and statefulness to flagd, which are not worth it. I think requiring a restart of flagd (and therefore the workload pod) is completely reasonable here. _Our goal is to support changing flag values without a restart, not sidecar configuration without a restart._ I think we might want to have the operator facilitate graceful rollout of this new config, but that's the extent of it. Cattle, not pets.

I think the naming of the CRD fields here is a bit tough. I actually like `provider/source`, but `source` seems a bit out of place as a way to mention what is actually a `FeatureFlagConfiguration`. I think `flagConfigruation` or `configuration` might make it clearer what kind of reference is happening here (see below):

```yaml
kind: FlagSourceConfiguration
metadata:
    name: end-to-end-fsc
spec:
    syncProviders:
    - source: end-to-end
      provider: kubernetes
    - source: end-to-end2      // should this be `configuration: end-to-end2   ?
      provider: kubernetes
```

james-milligan:
> * I couldnt seem to disable the cluster permissions with `openfeature.dev/allowkubernetessync: "false"`. Whether I had previously set this value to `"true"` or not seemed to have no impact. I was expecting that if I set this to `"false"`, I would see some permissions error in the flagd sidecar - I did not - in fact everything worked. Maybe I configured it incorrectly or I'm confused about the purpose of this annotation.

So the annotation is used internally to manage upgrades, when the manager container starts up it identifies all annotated pods and adds their permissions to the cluster role binding, ill update the documentation to reflect this + add a reference that its for internal usage only

> * Expected, I believe, but changes in `FlagSourceConfiguration` kinds had no impact until the workload was cycled - for instance, changing the syncProvider value from `kubernetes` => `filepath` didn't cause flagd to watch the filepath (more on this later, below).

> ### More throughts:
> I think that we should _certainly not_ support dynamically changing the mode of flagd operation when a `FlagSourceConfiguration` changes. I think this would add a lot of complexity and statefulness to flagd, which are not worth it. I think requiring a restart of flagd (and therefore the workload pod) is completely reasonable here. _Our goal is to support changing flag values without a restart, not sidecar configuration without a restart._ I think we might want to have the operator facilitate graceful rollout of this new config, but that's the extent of it. Cattle, not pets.

Without being a little 'hacky' we can't cleanly restart a single container in a pod, in the example described above the graceful rollout would require the bouncing of pods containing otherwise healthy containers. This being said the introduction of a new feature flag is more than likely going to arrive with the introduction of a code change in the accompanying pods, so I agree that a more destructive rollout is a perfectly fine option, rather than watching the configurations. The only exception to this is client facing flagd instances, in which the pod will be sat behind a sidecar with no accompanying pods, but I am against introducing 'context specific behaviour', and instead should manage this with proper rollout strategies.

> I think the naming of the CRD fields here is a bit tough. I actually like `provider/source`, but `source` seems a bit out of place as a way to mention what is actually a `FeatureFlagConfiguration`. I think `flagConfigruation` or `configuration` might make it clearer what kind of reference is happening here (see below):
> 
> ```yaml
> apiVersion: core.openfeature.dev/v1alpha3
> kind: FlagSourceConfiguration
> metadata:
>     name: end-to-end-fsc
> spec:
>     syncProviders:
>     - source: end-to-end
>       provider: kubernetes
>     - source: end-to-end2      // should this be `configuration: end-to-end2   ?
>       provider: kubernetes
> ```

I given some thought to this, the issue with configuration is it no longer fits with `host:port` sources, such as those with http (and soon grpc) syncs, perhaps we default to URI, matching the identifier we use internally in flagd?

skyerus:
I agree that `flagd` should not changes its mode of operation at runtime, the complexity far outweighs any benefits imo.

Crashing a container to rollout a mode of operation change is also less than ideal, I'm onboard for the operator facilitating the rollout as a workaround.

> I do however think we should mount the FeatureFlagConfiguration to allow for non-flagd sidecars to watch it if required, and also allow for us to introduce this behaviour to flagd if we see fit.

Could you elaborate a bit further on this? Mount it in which sense, as a config map volume? Introduce what behaviour to flagd?

I prefer `source` as the naming, configuration is more appropriate for CRDs but I think `source` sufficiently covers all cases.

james-milligan:
> > I do however think we should mount the FeatureFlagConfiguration to allow for non-flagd sidecars to watch it if required, and also allow for us to introduce this behaviour to flagd if we see fit.
> 
> Could you elaborate a bit further on this? Mount it in which sense, as a config map volume? Introduce what behaviour to flagd?
> 
Mounting the `FeatureFlagConfiguration` as a ConfigMap volume will allow for it to be ingested (if required) by the sidecar, and whilst this behaviour is not currently exhibited by flagd (we currently use env vars / args), it MAY exist in another feature flagging sidecar (and possibly also flagd in the future). 
It would also allow for an alternative to pulling configuration from environment variables in the application containers. This would also allow us to keep each value in its intended type, rather than converting to and from strings - which would be enforced by the validation webhook.

docs/annotations.md
```diff
@@ -11,32 +11,36 @@ Example:
...
+
 Example:
-```yaml
-    metadata:
+```
+  metadata:
     annotations:
         openfeature.dev/enabled: "true"
-        openfeature.dev/featureflagconfiguration: "demo, test/demo-2"
+        openfeature.dev/flagsourceconfiguration:"config-A, config-B"`
```
skyerus, line 25:
```suggestion
        openfeature.dev/flagsourceconfiguration:"config-A,config-B"`
```
Does spacing work? In the e2e tests we don't use spacing

docs/annotations.md
```diff
@@ -11,32 +11,36 @@ Example:
...
-If multiple CRs are provided, they are merged with the latest taking precedence, for example, in the scenario below, `config-B` will take priority in the merge, replacing duplicated values that are set in `config-A`.
-
 Example:
-```
-  metadata:
+```yaml
+    metadata:
     annotations:
         openfeature.dev/enabled: "true"
         openfeature.dev/featureflagconfiguration: "demo, test/demo-2"
```
skyerus, line 43:
```suggestion
        openfeature.dev/featureflagconfiguration: "demo,test/demo-2"
```

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
+# Flag Source configuration
+
+The injected sidecar is configured using the `FlagSourceConfiguration` CRD, the `openfeature.dev/flagsourceconfiguration` annotation is used to assign `Pods` with their respective `FlagSourceConfiguration` CRs. The annotation value is a comma separated list of values following one of 2 patterns: {NAME} or {NAMESPACE}/{NAME}. If no namespace is provided, it is assumed that the CR is within the same namespace as the deployed pod, for example:
+```
+    metadata:
+        namespace: test-ns
+        annotations:
+            openfeature.dev/enabled: "true"
+            openfeature.dev/flagsourceconfiguration:"config-A, test-ns-2/config-B"
```
skyerus, line 9:
```suggestion
            openfeature.dev/flagsourceconfiguration:"config-A,test-ns-2/config-B"
```

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+```
+In this example, 2 CRs are being used to configure the injected container (by default the the operator uses the `flagd:main` image), `config-A` (which is assumed to be in the namespace `test-ns`) and `config-B` from the `test-ns-2` namespace, with `config-B` taking precedence in the configuration merge.
+
+The `FlagSourceConfiguration` version `v1alpha3` CRD defines the a CR with the following example structure:
```
skyerus, line 13:
```suggestion
The `FlagSourceConfiguration` version `v1alpha3` CRD defines a CR with the following example structure:
```

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+    envVars:
+    - name: MY_ENV_VAR
+      value: my-env-value
+```
+
+The relevant `FlagSourceConfigurations` are passed to the operator by setting the `openfeature.dev/flagsourceconfiguration` annotation, and is responsible for providing the full configuration of the injected sidecar.
+
+## FlagSourceConfiguration Fields
+
+| Field      | Behavior | Type |
```
skyerus, line 42: It may be good to include a default column here

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+
+## SyncProvider Fields
+
+| Field      | Behavior | Type | 
+| ----------- | ----------- | ----------- |
+| Source      | Defines the URI of the flag source, this can be either a `host:port` or the `namespace/name` of a `FeatureFlagConfiguration`       | `string`       |
+| Provider      | Defines the provider to be used, can be set to `kubernetes`, `filepath` or `http`      | `string`       |
+| HttpSyncBearerToken      | Defines the bearer token to be used with a `http` sync. Has no effect if `Provider` is not `http`      | optional `string`      |
+
+Within the CRD there are 2 main objects, namely the `flagDSpec` and the `featureFlagSpec`, both offering a different set of configuration for the injected `flagd` sidecars.
```
skyerus, line 62: How does this pertain to the SyncProvider fields?

docs/getting_started.md
```diff
@@ -50,9 +50,25 @@ spec:
         targeting: {}
 ```
 
-### Reference the deployed FeatureFlagConfiguration within a Deployment spec annotation.
+### Reference the deployed FeatureFlagConfiguration within FlagSourceConfiguration.
 
-In this example, a`Deployment` containing a `busybox-curl` container is created. In the example below, the `metadata.annotations` object contains the required annotations for the operator to correctly configure and inject the `flagd` sidecar into each deployed `Pod`. The documentation for these annotations can be found [here](./annotations.md).
+The `FlagSourceConfiguration` defined below can be used to assign our `FeatureFlagConfiguration`, as well as any other configuration settings, to the injected sidecars. In this example we are also setting the port exposed by our injected container.
```
skyerus, line 55:
```suggestion
The `FlagSourceConfiguration` defined below can be used to assign the `FeatureFlagConfiguration`, as well as any other configuration settings, to the injected sidecars. In this example, the port exposed by the injected container is also set.
```

webhooks/pod_webhook.go
```diff
@@ -149,75 +112,236 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
-			m.Log.V(1).Info(fmt.Sprintf("failed to parse annotation %s error: %s", ffName, err.Error()))
-			return admission.Errored(http.StatusBadRequest, err)
-		}
-		// Check to see whether the FeatureFlagConfiguration has service or sync overrides
-		ff := m.getFeatureFlag(ctx, name, ns)
-		if reflect.DeepEqual(ff, corev1alpha1.FeatureFlagConfiguration{}) {
-			m.Log.V(1).Info(fmt.Sprintf("FeatureFlagConfiguration could not be found for %s", ffName))
-			return admission.Errored(http.StatusBadRequest, err)
+	// maintain backwards compatibility of the openfeature.dev/featureflagconfiguration annotation
+	ffConfigAnnotation, ffConfigAnnotationOk := pod.GetAnnotations()["openfeature.dev/featureflagconfiguration"]
```
skyerus, line 136: could benefit from `openfeature.dev/featureflagconfiguration` being a constant

webhooks/pod_webhook.go
```diff
@@ -149,75 +112,236 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+	}
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+}
+
+func (m *PodMutator) injectSidecar(
+	ctx context.Context,
+	pod *corev1.Pod,
+	flagSourceConfig *v1alpha1.FlagSourceConfigurationSpec,
+) ([]byte, error) {
+	m.Log.V(1).Info(fmt.Sprintf("Creating sidecar for pod %s/%s", pod.Namespace, pod.Name))
```
skyerus, line 159:
```suggestion
	m.Log.V(1).Info(fmt.Sprintf("creating sidecar for pod %s/%s", pod.Namespace, pod.Name))
```

webhooks/pod_webhook.go
```diff
@@ -149,75 +112,236 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+	if ff.Name == "" {
+		return fmt.Errorf("feature flag configuration %s/%s not found", ns, n)
 	}
-	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+	// add permissions to pod
+	if err := m.enableClusterRoleBinding(ctx, pod); err != nil {
+		return err
+	}
+	// mark pod with annotation (required to backfill permissions if they are dropped)
+	pod.Annotations["openfeature.dev/allowkubernetessync"] = "true"
```
skyerus, line 246: constant?

skyerus (commented):
Remove generated chart stuff

toddbaert:
> > > I do however think we should mount the FeatureFlagConfiguration to allow for non-flagd sidecars to watch it if required, and also allow for us to introduce this behaviour to flagd if we see fit.
> > 
> > 
> > Could you elaborate a bit further on this? Mount it in which sense, as a config map volume? Introduce what behaviour to flagd?
> 
> Mounting the `FeatureFlagConfiguration` as a ConfigMap volume will allow for it to be ingested (if required) by the sidecar, and whilst this behaviour is not currently exhibited by flagd (we currently use env vars / args), it MAY exist in another feature flagging sidecar (and possibly also flagd in the future). It would also allow for an alternative to pulling configuration from environment variables in the application containers. This would also allow us to keep each value in its intended type, rather than converting to and from strings - which would be enforced by the validation webhook.

I think you mean `FlagSourceConfiguration`?

docs/annotations.md
james-milligan: https://github.com/open-feature/open-feature-operator/blob/5f839209008ced69198cdf909659c0001e1b3910/webhooks/pod_webhook.go#L223

Spacing is handled in the parse func

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -146,6 +197,10 @@ func NewFlagSourceConfigurationSpec() (*FlagSourceConfigurationSpec, error) {
 		fsc.DefaultSyncProvider = SyncProviderType(syncProvider)
 	}
 
+	if envVarPrefix := os.Getenv(fmt.Sprintf("%s_%s", InputConfigurationEnvVarPrefix, SidecarEnvVarPrefix)); envVarPrefix != "" {
```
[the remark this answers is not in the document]
james-milligan: This was was originally following the existing pattern of all other keys, good catch

docs/flag_source_configuration.md
james-milligan: removed

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -171,83 +226,59 @@ func (fc *FlagSourceConfigurationSpec) Merge(new *FlagSourceConfigurationSpec) {
...
+			Name:  fmt.Sprintf("%s_%s", fc.EnvVarPrefix, envVar.Name),
+			Value: envVar.Value,
+		})
 	}
 
-	if fc.MetricsPort != defaultMetricPort {
+	if fc.MetricsPort != DefaultMetricPort {
 		envs = append(envs, corev1.EnvVar{
-			Name:  fmt.Sprintf("%s_%s", prefix, SidecarMetricPortEnvVar),
+			Name:  fmt.Sprintf("%s_%s", fc.EnvVarPrefix, SidecarMetricPortEnvVar),
```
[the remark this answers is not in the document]
james-milligan: The building of the key or 254 => 259

webhooks/pod_webhook.go
james-milligan: Introduced as individual constants (annotation prefix + suffix for each of the 3 annotations)

apis/core/v1alpha1/flagsourceconfiguration_types.go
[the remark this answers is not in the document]
skyerus: Building of all of these keys, they're effectively doing the same thing

config/overlays/helm/manager.yaml
```diff
@@ -34,6 +34,8 @@ spec:
             value: "{{ .Values.sidecarConfiguration.envVarPrefix }}"
           - name: SIDECAR_SYNC_PROVIDER
             value: "{{ .Values.sidecarConfiguration.defaultSyncProvider }}"
+            - name: SIDECAR_EVALUATOR
+            value: "{{ .Values.sidecarConfiguration.evaluator }}"
```
toddbaert, line 38: There's an indentation error here that is breaking helm packaging.

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -261,3 +293,11 @@ func (s SyncProviderType) IsHttp() bool {
 func (s SyncProviderType) IsFilepath() bool {
 	return s == SyncProviderFilepath
 }
+
+func envVarKey(prefix string, suffix string) string {
+	return fmt.Sprintf("%s_%s", prefix, suffix)
+}
```
toddbaert, line 299: gorgeous.

docs/annotations.md
```diff
@@ -11,32 +11,36 @@ Example:
...
+```
+  metadata:
     annotations:
         openfeature.dev/enabled: "true"
-        openfeature.dev/featureflagconfiguration: "demo, test/demo-2"
+        openfeature.dev/flagsourceconfiguration:"config-A, config-B"`
 ```
 
+### `openfeature.dev/allowkubernetessync`
+*This annotation is used internally by the operator.*
```
toddbaert, line 29: Thanks for making this clear.

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -92,60 +93,115 @@ type FlagSourceConfigurationSpec struct {
...
+	// +optional
+	EnvVars []corev1.EnvVar `json:"envVars"`
+
+	// EnvVarPrefix defines the prefix to be applied to all environment variables applied to the sidecar, default FLAGD
+	EnvVarPrefix string `json:"envVarPrefix"`
+}
+
+type SyncProvider struct {
+	Source   string           `json:"source"`
+	Provider SyncProviderType `json:"provider"`
```
toddbaert, line 112: Is there value in making the `SyncProviderType` optional and falling back to the default if it's not set?

The [`syncProviders` field](https://github.com/open-feature/open-feature-operator/pull/344/files#diff-b666c9521abcc0b8e6e7adf5bd7fa76ddbc67a97e44d05079c1e3b6f2eb10857R99) would still have a use in aggregating multiple sources to a single `FlagSourceConfiguration`. However, if we make this change, I think `sources` might be a better name for the field.

toddbaert (changes requested):
Looks good to me [pending one yaml formatting fix](https://github.com/open-feature/open-feature-operator/pull/344#discussion_r1106171333). Tested locally, trying to break it every way I could, and it seems good. I'd like some feedback on [this idea](https://github.com/open-feature/open-feature-operator/pull/344#discussion_r1106246201), but it's not blocking my approval.

config/overlays/helm/manager.yaml
james-milligan: resolved

apis/core/v1alpha1/flagsourceconfiguration_types.go
james-milligan: In my opinion the `DefaultSyncProvider` configuration presented more value in versions `v1alpha1` and `v1alpha2` where the configuration was a little more convoluted, this being said I am not against reintroducing this field.
I would be keen to introduce this change in a second PR however, as this would require updates to the validation webhook for the `FlagSourceConfugration` CR (ensure that a default sync provider is set on the resource if any of the SyncProviders / Sources are missing the provider field), and I am conscious this PR is already bloated.

skyerus: I think simplifying the process to allow users to only supply sources (with the sync provider falling back to the default) makes sense if the behaviour is well documented.
I agree that renaming to `sources` makes more sense if this approach is taken.
In which case, it makes sense to do it in this release while the behaviour is already being altered and before the naming is final.

toddbaert: I agree this can be done later if at all.

james-milligan: Ive introduced these changes following Skyes justification - once the changes re introduced we would need a new version to change the name!

toddbaert (approved):
Approved, but you got conflict bombed.

toddbaert:
@james-milligan sorry to do this to you, but I'm merging https://github.com/open-feature/open-feature-operator/pull/345, which will create some conflicts... however, they are all easy to fix by simply running `make helm-package` after rebasing-on/merging main, which will regenerate the charts (individually, instead of at `rendered.yaml`.

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -95,68 +95,128 @@ type FlagSourceConfigurationSpec struct {
 	// +optional
 	DefaultSyncProvider SyncProviderType `json:"defaultSyncProvider"`
 
+	// Sources defines the syncProviders and associated configuration to be applied to the sidecar
+	// +optional
+	Sources []Source `json:"sources"`
```
toddbaert, line 100: We may want to make this field required, and also force at least 1 item in the array. As is, if you don't put anything here, flagd attempts to start with no args and it results in a CrashLoopBackoff ( @beeme1mr  found this in his testing).

`// +kubebuilder:validation:MinItems` seems to exist for this purpose, see [here](https://book.kubebuilder.io/reference/markers/crd-validation.html).

toddbaert (approved):
Re-tested again against latest changes with @beeme1mr . Only new finding was [this](https://github.com/open-feature/open-feature-operator/pull/344/files#r1108837821).

apis/core/v1alpha1/flagsourceconfiguration_types.go
beeme1mr: The unique validator may also make sense.

config/samples/core_v1alpha3_flagsourceconfiguration.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/name: flagsourceconfiguration
+    app.kubernetes.io/instance: flagsourceconfiguration-sample
+    app.kubernetes.io/part-of: open-feature-operator
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: open-feature-operator
+  name: flagsourceconfiguration-sample
+spec:
+  # TODO(user): Add fields here
```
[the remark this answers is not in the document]
james-milligan: Added a sample to the end to end deployment

apis/core/v1alpha1/flagsourceconfiguration_types.go
james-milligan: Ive introduced the `MinItems` tag, however, adding the unique validator cannot also be added as `the runtime complexity becomes quadratic`. If we want to introduce this change we can always introduce the logic into the validation webhook

controllers/flagsourceconfiguration_controller.go
```diff
@@ -18,45 +18,146 @@ package controllers
...
+		annotations := deployment.Spec.Template.Annotations
+		annotation, ok := annotations[fmt.Sprintf("%s/%s", OpenFeatureAnnotationRoot, FlagSourceConfigurationAnnotation)]
+		if !ok {
+			continue
+		}
+		if isUsingConfiguration(fsConfig.Namespace, fsConfig.Name, deployment.Namespace, annotation) {
+			r.Log.Info(fmt.Sprintf("restarting deployment %s/%s", deployment.Namespace, deployment.Name))
+			deployment.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"] = time.Now().Format(time.RFC3339)
+			if err := r.Client.Update(ctx, &deployment); err != nil {
+				r.Log.V(1).Error(err, fmt.Sprintf("Failed to update Deployment: %s", err.Error()))
```
skyerus, line 103: Supplying the error twice is unnecessary. It might be good to include which deployment in the message however

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+
+The injected sidecar is configured using the `FlagSourceConfiguration` CRD, the `openfeature.dev/flagsourceconfiguration` annotation is used to assign `Pods` with their respective `FlagSourceConfiguration` CRs. The annotation value is a comma separated list of values following one of 2 patterns: {NAME} or {NAMESPACE}/{NAME}. If no namespace is provided, it is assumed that the CR is within the same namespace as the deployed pod, for example:
+```
+    metadata:
+        namespace: test-ns
+        annotations:
+            openfeature.dev/enabled: "true"
+            openfeature.dev/flagsourceconfiguration:"config-A, test-ns-2/config-B"
+```
+In this example, 2 CRs are being used to configure the injected container (by default the the operator uses the `flagd:main` image), `config-A` (which is assumed to be in the namespace `test-ns`) and `config-B` from the `test-ns-2` namespace, with `config-B` taking precedence in the configuration merge.
```
skyerus, line 11: Is the configuration merging priority system defined outside of this example?

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+
+| Field | Behavior | Type | Default | 
+| ----------- | ----------- | ----------- | ----------- |
+| MetricsPort | Defines the port for flagd to serve metrics on | optional `int32`| `8013` |
+| Port   | Defines the port for flagd to listen on | optional `int32` | `8014` |
+| SocketPath   | Defines the unix socket path to listen on        | optional `string`       | `""` |
+| SyncProviderArgs   | String arguments passed to the sidecar on startup, flagd documentation can be found [here](https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md)        | optional `array of strings`, key values separated by `=`, e.g `key=value`       | `""` | 
+| Image   | Allows for the sidecar image to be overridden        | optional `string`       | `ghcr.io/open-feature/flagd` | 
+| Tag   |  Tag to be appended to the sidecar image | optional `string`       | `main` |
+| Sources   |  An array of objects defining configuration and sources for each sync provider to use within flagd, object documentation can be found below        | optional `array of objects`       |`[]` |
```
skyerus, line 50:
```suggestion
| Sources   |  An array of objects defining configuration and sources for each sync provider to use within flagd, documentation of the object is directly below this table        | optional `array of objects`       |`[]` |
```

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+| EnvVars   |  An array of environment variables to be applied to the sidecar, all names will be prepended with the EnvVarPrefix    | optional `array of environment variables`       | `[]` |
```
skyerus, line 51:
```suggestion
| EnvVars   |  An array of environment variables to be applied to the sidecar, all names become prepended with the EnvVarPrefix    | optional `array of environment variables`       | `[]` | 
```

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+
+| Field      | Behavior | Type | 
+| ----------- | ----------- | ----------- |
+| Source      | Defines the URI of the flag source, this can be either a `host:port` or the `namespace/name` of a `FeatureFlagConfiguration`       | `string`       |
+| Provider      | Defines the provider to be used, can be set to `kubernetes`, `filepath` or `http`. If not provided the default sync provider is used.     | optional `string`       |
+| HttpSyncBearerToken      | Defines the bearer token to be used with a `http` sync. Has no effect if `Provider` is not `http`      | optional `string`      |
+
+## Configuration Merging
+
+When multiple `FlagSourceConfigurations` are passed the configurations are merged, the last `CR` will take precedence over the first, with any configuration from the deprecated `FlagDSpec` field of the `FeatureFlagConfiguration` CRD taking the lowest priority.
```
skyerus, line 66: A mermaid diagram of this merge priority system could be helpful (example: https://github.com/open-feature/flagd/blob/main/docs/other_resources/creating_providers.md#configuration)

skyerus, line 66:
```suggestion
When multiple `FlagSourceConfigurations` are provided, the configurations are merged. The last `CR` takes precedence over the first, with any configuration from the deprecated `FlagDSpec` field of the `FeatureFlagConfiguration` CRD taking the lowest priority. 
```

docs/flag_source_configuration.md
```diff
@@ -0,0 +1,100 @@
...
+| Field      | Behavior | Type | 
+| ----------- | ----------- | ----------- |
+| Source      | Defines the URI of the flag source, this can be either a `host:port` or the `namespace/name` of a `FeatureFlagConfiguration`       | `string`       |
+| Provider      | Defines the provider to be used, can be set to `kubernetes`, `filepath` or `http`. If not provided the default sync provider is used.     | optional `string`       |
+| HttpSyncBearerToken      | Defines the bearer token to be used with a `http` sync. Has no effect if `Provider` is not `http`      | optional `string`      |
+
+## Configuration Merging
+
+When multiple `FlagSourceConfigurations` are passed the configurations are merged, the last `CR` will take precedence over the first, with any configuration from the deprecated `FlagDSpec` field of the `FeatureFlagConfiguration` CRD taking the lowest priority. 
+An example of this behavior can be found below:
```
skyerus, line 67:
```suggestion
An example of this behavior:
```

docs/getting_started.md
```diff
@@ -50,9 +50,25 @@ spec:
...
+spec:
+  syncProviders:
+  - source: featureflagconfiguration-sample
+    provider: kubernetes
+  port: 8080
+```
+
+### Reference the deployed FlagSourceConfiguration within a Deployment spec annotation.
+
+In this example, a `Deployment` containing a `busybox-curl` container is created. In the example below, the `metadata.annotations` object contains the required annotations for the operator to correctly configure and inject the `flagd` sidecar into each deployed `Pod`. The documentation for these annotations can be found [here](./annotations.md).
```
skyerus, line 71:
```suggestion
In this example, a `Deployment` containing a `busybox-curl` container is created. In the configuration below, the `metadata.annotations` object contains the required annotations for the operator to correctly configure and inject the `flagd` sidecar into each deployed `Pod`. The documentation for these annotations can be found [here](./annotations.md).
```

docs/flag_source_configuration.md
james-milligan: Yes, there is a configuration merging section below

controllers/flagsourceconfiguration_controller.go
james-milligan: resolved

docs/flag_source_configuration.md
james-milligan: included a diagram

skyerus (approved):

apis/core/v1alpha2/flagsourceconfiguration_types.go
```diff
@@ -63,6 +63,15 @@ type FlagSourceConfigurationSpec struct {
 	// LogFormat allows for the sidecar log format to be overridden, defaults to 'json'
 	// +optional
 	LogFormat string `json:"logFormat"`
+
```
odubajDT, line 66: Also here, what is the reason why there are changes in v1alpha2 API version? AFAIK once the API version is released, no changes should be made to this resource definition. If there are changes required, the new version should be introduced.

controllers/controllers_test.go
```diff
@@ -0,0 +1,104 @@
+package controllers
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/open-feature/open-feature-operator/apis/core/v1alpha1"
```
odubajDT, line 8: I would expect to use v1alpha3 version here

controllers/flagsourceconfiguration_controller.go
```diff
@@ -18,45 +18,146 @@ package controllers
...
+	appsV1 "k8s.io/api/apps/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
+	"github.com/go-logr/logr"
 	corev1alpha1 "github.com/open-feature/open-feature-operator/apis/core/v1alpha1"
```
odubajDT, line 34: all these imports need to be adapted to v1alpha3, v1alpha1 and v1alpha2 versions can be present and used only in conversion webhooks

main.go
```diff
@@ -38,8 +39,10 @@ import (
 
 	corev1alpha1 "github.com/open-feature/open-feature-operator/apis/core/v1alpha1"
 	corev1alpha2 "github.com/open-feature/open-feature-operator/apis/core/v1alpha2"
+	corev1alpha3 "github.com/open-feature/open-feature-operator/apis/core/v1alpha3"
```
odubajDT, line 42: only the latest v1alpha3 version should be used here

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -95,68 +95,134 @@ type FlagSourceConfigurationSpec struct {
 	// +optional
 	DefaultSyncProvider SyncProviderType `json:"defaultSyncProvider"`
 
+	// Sources defines the syncProviders and associated configuration to be applied to the sidecar
```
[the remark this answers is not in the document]
toddbaert: I think the answer is simply that we use v1 as our hub, only because it was easier/faster. This is a non-breaking change so it doesn't cause a problem AFAIK

For alpha it was the path of least resistance. Eventually to keep things cleaner, when there is less flux in our CRDs, we should probably make beta the hub and updated all the conversions. I don't think that should be done in this PR though.

controllers/flagsourceconfiguration_controller.go
toddbaert: See [here](https://github.com/open-feature/open-feature-operator/pull/344/files#r1113202711)

.gitignore
```diff
@@ -1,3 +1,8 @@
+# ignore generated templates
+templates/*.yaml
+# templates/namespace.yaml is not generated, it's a special case
+!templates/namespace.yaml
+
```
toddbaert, line 5: This looks similar to the `.gitignore` lines inside the the [charts](https://github.com/open-feature/open-feature-operator/blob/main/chart/open-feature-operator/.gitignore)

Are they needed? Why would there be a `/template` dir here?

apis/core/v1alpha1/flagsourceconfiguration_types.go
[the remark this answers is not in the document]
james-milligan: @odubajDT if you're happy to take this approach, continuing along the path of least resistance followed by a 'cleanup' in the beta release then i'm happy to merge this in now

.gitignore
james-milligan: Looks like I confused myself on the latest rounds of conflicts, I have resolved this

apis/core/v1alpha1/flagsourceconfiguration_types.go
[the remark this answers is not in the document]
toddbaert: Here is the existing research issue for this: https://github.com/open-feature/open-feature-operator/issues/352

beeme1mr (approved):

apis/core/v1alpha2/flagsourceconfiguration_types.go
james-milligan: ive reverted these changes

PR #359, opened 2023-02-21
chore: adapt Makefile for local dev deployment + introduce report logs for e2e-tests

odubajDT:
## This PR

- adapts Makefile to for local dev deployment. Currently, you should be able to deploy the local version of your operator just with the single command to your k8s cluster. Please be aware, it is using the cluster your context is pointing to.

Example:

`make build-deploy-operator TAG=main RELEASE_REGISTRY=docker.io/odubajdt`

Also, this PR introduces a report logs for e2e-tests -> there is possible to see a state of the resources in the test cluster after finishing the tests


toddbaert:
Wow, this is a very nice thing to have. I will give this a test later! Thanks!

Makefile
```diff
@@ -1,6 +1,8 @@
-
-# Image URL to use all building/pushing image targets
-IMG ?= controller:latest
+RELEASE_REGISTRY?=ghcr.io/openfeature
+TAG?=latest
+RELEASE_IMAGE?=operator:$(TAG)
```
toddbaert, line 3: I think it might be a _bit_ less confusing if we do something like:

```
RELEASE_IMAGE?=$(RELEASE_NAME):$(TAG)
```

Then I can control everything like:

```
make build-deploy-operator TAG=mytag RELEASE_REGISTRY=toddbaert RELEASE_NAME=myImgName
```

Which will pull `toddbaert/myImgName:mytag`

But this is just my opinion! I can go with the current way as well.

toddbaert:
This works great and looks good to me. Definitely a very nice time-saver! I suggest that you add a bit about this to `CONTRIBUTING.md`!

My only (optional) suggestion is [this](https://github.com/open-feature/open-feature-operator/pull/359#discussion_r1113359651).

toddbaert (approved):

Makefile
odubajDT: sure, we can adapt it, thanks for the feedback :)

odubajDT: adapted

Issue #354, opened 2023-02-15
[docs] Update docs to mention separate CRD install


toddbaert:
Document how to install CRDs separately in case of helm, similar to https://cert-manager.io/docs/installation/helm/.

See https://github.com/open-feature/open-feature-operator/pull/345

PR #361, opened 2023-02-22
chore: document crd upgrades

toddbaert:
Adds docs for upgrading CRDs via helm.

Fixes: https://github.com/open-feature/open-feature-operator/issues/354

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -19,44 +19,40 @@ Install the latest helm release:
 ```sh
 helm repo add openfeature https://open-feature.github.io/open-feature-operator/
 helm repo update
-helm upgrade -i openfeature openfeature/open-feature-operator
+helm upgrade --install openfeature openfeature/open-feature-operator
 ```
 ### Upgrading
 
 ```sh
-helm upgrade -i openfeature openfeature/open-feature-operator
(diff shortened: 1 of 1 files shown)


Kavindu-Dodan (approved):

thisthat (approved):
LGTM

mowies (commented):
some general things that may help with UX:
- For docs I think it always makes sense to use the full name flags and not shorthand, so that command are easier to read for users (i.e. `-i` -> `--install`)
- you can call `helm template` also on remote charts then you could skip the download step

toddbaert:
> * For docs I think it always makes sense to use the full name flags and not shorthand, so that command are easier to read for users (i.e. `-i` -> `--install`)

I actually strongly agree with this. This is usually my policy in scripts too, to keep things readable. I will make this change.

> * you can call `helm template` also on remote charts then you could skip the download step

TIL! I will update to this, since it's simpler I think.

toddbaert:
@mowies I have done the needful, please have a quick look. Thanks for those suggestions.

toddbaert:
@mowies I'm going to merge this since I _think_ I've addressed your concerns and I'd like to have this documented for our next release. If that's not true, please comment and I will resolve with a new PR.

PR #369, opened 2023-03-06
fix: refactor + move hub to v1alpha3

odubajDT:
## This PR

- refactoring
- move API hub v1alpha3
- adapt conversions


odubajDT:
> @odubajDT seems this had gone stale, hence shall we close this?

I guess yes, we will probably move the HUB version to beta, when we will decide for it

Issue #370, opened 2023-03-06
refactoring: open-feature-operator


odubajDT:
## Goal

The goal of this ticket is to refactor the operator code and improve the test coverage

### What should be done
- define how should resources in different API versions look like and make a standard from it -> currently a lot of new fields were introduced to v1alpha1 to maintain backwards compatibility (AFAIK this can be achieved by moving the HUB to v1alpha3) -> In this case, I would prefer to take the state of the CRDs from the latest release version, so it will be standardized since that
- remove unneeded conversion webhooks -> currently having conversion webhook for each API version and resource, it is enough to have only one conversion webhook registered for the HUB version, spikes do not need their own webhooks
- make v1alpha3 HUB version -> v1alpha3 contains all of the information that are in v1alpha1 or v1alpha2 (but they do not contain all information that is present in v1alpha3), therefore it is easy to convert between versions to v1alpha3 without loosing data
- use v1alpha3 as imported API version in controllers
- move testing of conversion webhooks from using ginkgo to standard unit tests (potentially also integration tests with the use of kuttl)
- use renovate to update versions of flagD in the operator instead of external shell scripts
- introduce integration tests using kuttl for basic use-case scenarios

A first draft PR containing some the changes (point 2, 3, 4) is available here: https://github.com/open-feature/open-feature-operator/pull/369

bacherfl:
A PoC for using Kuttl as a e2e test framework is available here: https://github.com/bacherfl/open-feature-operator/pull/1

odubajDT:
> A PoC for using Kuttl as a e2e test framework is available here: [bacherfl#1](https://github.com/bacherfl/open-feature-operator/pull/1)

Awesome job! Is there a reason why this PR should be just a PoC and cannot be directly merged into the repository ? @bacherfl It seems like with some minor polishing, we can directly use that in the repo as part of the CI pipeline

bacherfl:
> > A PoC for using Kuttl as a e2e test framework is available here: [bacherfl#1](https://github.com/bacherfl/open-feature-operator/pull/1)
> 
> Awesome job! Is there a reason why this PR should be just a PoC and cannot be directly merged into the repository ? @bacherfl It seems like with some minor polishing, we can directly use that in the repo as part of the CI pipeline

With some additional polishing (removing unneeded files and commands, etc.), the PR should be good to go - I have created a separate issue for this task here: https://github.com/open-feature/open-feature-operator/issues/391

[...]

PR #385, opened 2023-03-14
chore: remove unneeded conversion webhooks + introduce unit tests for conversion functions

odubajDT:
## This PR

- removes unneeded conversion webhooks from spike versions
- adapt conversion functions
- introduces unit tests for conversion functions
- removes conversion webhook tests from ginkgo testsuite
- bumps ginkgo version to v2
- introduce codecov.yaml for configuring test-coverage computations

### Related Issues

Part of #370


go.mod
```diff
@@ -86,3 +85,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.2.3 // indirect
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
+
+replace golang.org/x/net => golang.org/x/net v0.7.0
```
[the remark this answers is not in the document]
odubajDT: yes, there is a vulnerability in v0.5.0, therefore it was needed

.github/workflows/pr-checks.yml
```diff
@@ -71,6 +71,41 @@ jobs:
         run: make test
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
+        with:
+          flags: tests
+
+  unit-test:
```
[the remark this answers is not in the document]
odubajDT: unit-tests is a separate job, but the test-coverage of the code is computed together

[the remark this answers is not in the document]
skyerus: Unless I'm mistaken, `make test` covers all tests whereas `make unit-test` covers a subset of the tests (under the apis dir). My concern is that if more unit tests are introduced in a different dir they won't be covered by this job. The pattern I've seen elsewhere is to run the unit tests with the test `-short` flag and check the flag in non-unit tests - https://github.com/open-feature/flagd/blob/main/flagd/tests/integration/caching_test.go#L20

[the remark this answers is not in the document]
odubajDT: adapted

Makefile
```diff
@@ -67,7 +67,17 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
```
skyerus, line 69: Is the `test` name still appropriate given it's now only targeting controllers/webhooks? 🤔 
I'm not sure what appropriate naming is best here but I think the nuances of `make test` vs `make unit-test` may cause confusion

odubajDT: sure, we can change it, `unit-test` and `component-test` might be good?

skyerus: That sounds good. Maybe `test` can run both?

odubajDT: I would rather keep it separated, it's not that big of a deal to run 2 make commands

skyerus: Fine by me

odubajDT: adapted

Kavindu-Dodan (approved):

Makefile
```diff
@@ -65,9 +65,19 @@ fmt: ## Run go fmt against code.
 vet: ## Run go vet against code.
 	go vet ./...
 
-.PHONY: test
+.PHONY: component-test
 test: manifests generate fmt vet envtest ## Run tests.
```
skyerus, line 69:
```suggestion
component-test: manifests generate fmt vet envtest ## Run tests.
```

skyerus (approved):

PR #387, opened 2023-03-14
feat: client side support

### LINKED ISSUE #371: [OFO/flagd] Research and document flagd client support
With client [spec changes](https://github.com/open-feature/spec/issues/167) and [SDKs](https://github.com/open-feature/js-sdk/issues/367) on the way, we need to decide what sort of support we want for the client/web in OFO (if any). We already have the required bulk-resolution functionality in flagd, as well as CORs options to support local development

@james-milligan already did some research here with a PoC deployment, I believe. That may feed into this discussion.

If we were to support this, we'd want enhancements to the operator that could facilitate the following:

- Flagd must be exposed via a public ingress (handled by OFO or just documented)
- Flagd must support a bi-directional HTTP2 connections
- A client must be able to access flagd
- Flagd must be horizontally scalable based on load or number of connections

Non functionally, we should consider:

- updates to the [threat model](https://github.com/open-feature/open-feature-operator/blob/main/docs/threat_model.md)
- load considerations: How many persistent connection can be supported?

skyerus:
## This PR

- WIP of client side support

### Related Issues

Fixes #371


beeme1mr:
We've decided to not provide client-side deployment abstractions.

PR #390, opened 2023-03-14
enable flagd probes

Kavindu-Dodan:
## This PR
Relates to #408

Introduce flagd Liveness and Readiness probes for side-car deployed through OFO. Default is `enabled` (probes are enabled).

Behavior can be controlled through `FlagSourceConfiguration`'s field `probesEnabled` .

`probesEnabled`: `true`, `false`

Consider following configurations, 

Probes disabled, 

```yaml
kind: FlagSourceConfiguration
metadata:
  name: myappfs
spec:
  probesEnabled: false #probe disabled
  sources:
  - source: flags/flags
    provider: kubernetes
```

Probes enabled (default),

```yaml
kind: FlagSourceConfiguration
metadata:
  name: myappfs
spec:
  sources:
  - source: flags/flags
    provider: kubernetes
```


beeme1mr:
Hey @Kavindu-Dodan, could you please update the Artifact Hub readme with the new information?

https://github.com/open-feature/open-feature-operator/blob/main/chart/open-feature-operator/README.md#sidecar-configuration

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -175,6 +179,10 @@ func NewFlagSourceConfigurationSpec() (*FlagSourceConfigurationSpec, error) {
 		RolloutOnChange:     nil,
 	}
 
+	// set default value derived from constant default
+	probes := defaultProbesEnabled
+	fsc.ProbesEnabled = &probes
```
thisthat, line 184: We could also use the `// +kubebuilder:default` comment in the Spec definition at line 122. 
https://book.kubebuilder.io/reference/markers/crd-validation.html

thisthat: Adding fields to a previous version feels weird tho

Kavindu-Dodan: Seems this fails as I rely on a pointer to avoid default being set in CRD (defaulting to false if unset, which I want to avoid)

Kavindu-Dodan:
> Hey @Kavindu-Dodan, could you please update the Artifact Hub readme with the new information?
> 
> https://github.com/open-feature/open-feature-operator/blob/main/chart/open-feature-operator/README.md#sidecar-configuration

Thanks, I updated (along with formatting improvement) the readme. Also, your review helped to validate Helm configurations which I missed :)

beeme1mr (approved):
Looks good from a feature and documentation perspective. Nice job with the PR description by the way!

thisthat (approved):

james-milligan (approved):

apis/core/v1alpha1/flagsourceconfiguration_types.go
odubajDT: I guess we should keep this approach for v1alpha* versions but move away from it when moving to v1beta*

Kavindu-Dodan: Agree :) And refactoring right now is overkill.

PR #404, opened 2023-03-17
chore: split controllers to separate packages + cover them with unit tests

### LINKED ISSUE #393: Move controllers into their own packages
# Short abstract

Right now, all controllers (FeatureFlagConfigurationController, FlagSourceConfigurationController) are located in a common controller package. With this issue I would like to propose to move each of those controllers into their own package.

# Why 

In general, packages should be kept as small as possible to make them easy to understand. Also, in the ginkgo test suite, the controllers might interfere with each other when new tests are added. We recently observed that behaviour in https://github.com/keptn/lifecycle-toolkit, where the tests were failing randomly. Moving the controllers to their own packages will help to isolate them from each other when running the test suites

# Tasks

- [ ] Create a new package for each controller, i.e. `controllers/featureflagconfiguration/controller.go` and `controllers/flagsourceconfiguration/controller.go`
- [ ] Code that is shared between the existing controllers should be moved into a common package, e.g. `controllers/common/common.go`

odubajDT:
## This PR

- moves controllers to separate packages
- moves common functions/constants into separate common package
- migrates ginkgo controller tests to basic unit tests

### Related Issues

Fixes #393


AlexsJones:
Can you tell us why this is a good idea perhaps? I've not touched this code in a few months and the delta seems large.

odubajDT:
This PR is going to be split into 2 due to the size

odubajDT:
Split of the PR completed, the follow-up PR with additional refactoring and unit tests will follow up

controllers/core/featureflagconfiguration/controller_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+				cm2.OwnerReferences = append(cm2.OwnerReferences, v1alpha1.GetFfReference(ffConfig2))
+				err := r.Client.Update(ctx, cm2)
+				require.Nil(t, err)
+			}
+
+			_, err = r.Reconcile(ctx, req)
+			require.Nil(t, err)
+
+			ffConfig2 := &v1alpha1.FeatureFlagConfiguration{}
```
skyerus, line 130: Is there more appropriate var naming than ffConfig2?

odubajDT: fixed

skyerus (approved):
I like the splitting of the controllers package, nice

odubajDT:
> I like the splitting of the controllers package, nice

Thank you!

odubajDT:
> I'll give this a thorough review tomorrow!

Thank you!

apis/core/v1alpha1/featureflagconfiguration_types.go
```diff
@@ -145,3 +145,10 @@ func FeatureFlagConfigurationId(namespace, name string) string {
 func FeatureFlagConfigurationConfigMapKey(namespace, name string) string {
 	return fmt.Sprintf("%s.flagd.json", FeatureFlagConfigurationId(namespace, name))
 }
+
+func (p *FeatureFlagServiceProvider) IsSet() bool {
+	if p == nil || p.Name == "" {
+		return false
+	}
+	return true
```
toddbaert, line 153:
```suggestion
	return p != null && p != ""
```

controllers/core/flagsourceconfiguration/controller_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+			_, err = r.Reconcile(ctx, req)
+			require.Nil(t, err)
+
+			if tt.deployment != nil {
+				deployment := &appsv1.Deployment{}
+				err = fakeClient.Get(ctx, types.NamespacedName{Name: deploymentName, Namespace: testNamespace}, deployment)
+				require.Nil(t, err)
+
+				require.Equal(t, tt.restarted2, deployment.Spec.Template.ObjectMeta.Annotations["kubectl.kubernetes.io/restartedAt"])
```
toddbaert, line 102: Is there a chance that this assertion fails if things are really slow or these tests are reordered? Might it be better to instead test that `kubectl.kubernetes.io/restartedAt` is greater than `restarted2`? Maybe I'm confused abou twhat this test is doing.

toddbaert (approved):
Approved with some minor comments, questions.

controllers/core/flagsourceconfiguration/controller_test.go
odubajDT: The tests can be reordered and everything should work as expected. Each Iteration is completely independent from the other as we are using `t.Run()`, which is basically a new test with clean context.

 If the test is slow, hard to say to be honest, maybe worth merging and if we have some problems or flakyness in the future, we can always adapt it. Would this work ?

odubajDT: sure, would better description/comments do the trick for you ?

Kavindu-Dodan: @odubajDT  IMO yes. May be comment explaining what we do here and better naming for `restarted1` & `restarted2`

odubajDT: fixed :+1:

Kavindu-Dodan (approved):

controllers/core/flagsourceconfiguration/controller_test.go
Kavindu-Dodan: @toddbaert I think this is fixed. Can you also check ? :)

james-milligan (approved):

Issue #391, opened 2023-03-15
 Implement e2e tests in OF operator using KUTTL


bacherfl:
# Short Abstract 

Currently, the E2E tests are using a set of shell scripts that apply a FeatureFlag resource and then retrieve the value of a feature flag via the flagd API to evaluate the response.
These tests could potentially benefit from using KUTTL (https://kuttl.dev/), with which we have made good experiences in KLT (see https://github.com/keptn/lifecycle-toolkit/tree/main/test/integration).

# Why 

This framework allows to define test cases using Kubernetes manifests that are applied in one step, and then be evaluated using assertions that check for certain conditions being available on kubernetes resources (e.g. the status of a deployment). The framework will also take care of tasks such as isolating test scenarios from each other by running them in separate namespaces, as well as implementing the retry/timeout logic for the evaluation of the assertions.

The validation steps that are currently done via bash scripts can be reused by executing those in a K8s Job - Kuttl can then validate the status of that job.

Also, KUTTL may open the door for more complex e2e test scenarios with an annotated deployment being applied and then asserting if the flagd sidecar has successfully been injected

# Tasks

- [ ] Migrate the existing e2e test to Kuttl
- [ ] Add a command to the Makefile to execute these tests
- [ ] Execute these tests in the pipeline

# References

A proof of concept for this can be found in the following PR: https://github.com/bacherfl/open-feature-operator/pull/1 - this PR already wires up everything that is needed to also execute the tests in the pipeline, so with some further polishing it should be in a state in which it can be merged

RealAnna:
I would like to take this one!

PR #410, opened 2023-03-21
chore: refactor pod webhook mutator

RealAnna:
## This PR

makes some refactoring to prepare for kuttl tests implementation
- refactor pod mutator to reduce the complexity of the handler function 

### Related Issues

part of #391

### DIFF
--- a/webhooks/pod_webhook.go
+++ b/webhooks/pod_webhook.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"k8s.io/apimachinery/pkg/util/intstr"
 	"net/http"
 	"reflect"
 	"strings"
 	"time"
+
+	"k8s.io/apimachinery/pkg/util/intstr"
 
 	goErr "errors"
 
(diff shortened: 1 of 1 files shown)


webhooks/pod_webhook.go
```diff
@@ -118,22 +99,43 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
+}
+
+func (m *PodMutator) createFSConfigSpec(ctx context.Context, req admission.Request, annotations map[string]string, pod *corev1.Pod) (*v1alpha1.FlagSourceConfigurationSpec, int32, error) {
+	// Check configuration
+	fscNames := []string{}
+	val, ok := annotations[fmt.Sprintf("%s/%s", OpenFeatureAnnotationPrefix, FlagSourceConfigurationAnnotation)]
+	if ok {
+		fscNames = parseList(val)
+	}
```
toddbaert, line 121: We're doing this check in 2 places now. Does it make more sense to keep it early?

Kavindu-Dodan (approved):

webhooks/pod_webhook.go
toddbaert: Agreed!

beeme1mr (approved):

PR #411, opened 2023-03-22
feat(test): substitute kuttl to bash e2e test

### LINKED ISSUE #391: Implement e2e tests in OF operator using KUTTL

RealAnna:
## This PR

adds two e2e tests:
- one to check if flagd is injected and if the flags are available
- the other checks if we do not inject flagd in case the deployment has it disabled

### Related Issues

Fixes #391

## Notes
should we get rid of the scripts after migration?


config/manager/kustomization.yaml
```diff
@@ -12,5 +10,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: docker.io/annadreal/ofo
```
skyerus, line 13: accidental check in

skyerus (commented):
LGTM other than the accidental check in, nice :)

odubajDT:
part of #370

PR #412, opened 2023-03-22
feat: kube-flagd-proxy deployment

james-milligan:
## This PR

- adds new provider type `kube-proxy`
```
  sources:
  - source: default/example
    provider: kube-proxy
```
Use the `kube-proxy` provider + create a deployment with the normal deployment path, the operator will deploy Kubernetes proxy, denying pod deployment until it is ready to receive requests 

### Related Issues

Fixes #1234523


config/rbac/kustomization.yaml
```diff
@@ -4,7 +4,8 @@ resources:
 # if your manager will use a service account that exists at
 # runtime. Be sure to update RoleBinding and ClusterRoleBinding
 # subjects if changing service account names.
-- service_account.yaml
+- service_account_kube_proxy.yaml
```
beeme1mr, line 7: Will a Kubernetes admin need to do anything special when upgrading to this version of OFO?

webhooks/pod_webhook.go
```diff
@@ -230,6 +257,146 @@ func (m *PodMutator) injectSidecar(
...
+	configB, err := json.Marshal(config)
+	if err != nil {
+		return err
+	}
+
+	sidecar.Args = append(
+		sidecar.Args,
+		"--sources",
+		string(configB),
+		"--debug",
```
beeme1mr, line 395: Should debugging be enabled by default?

webhooks/pod_webhook.go
```diff
@@ -39,11 +43,22 @@ const (
 	ProbeReadiness                                       = "/readyz"
 	ProbeLiveness                                        = "/healthz"
 	ProbeInitialDelay                                    = 5
+	kubeProxyDeploymentName                              = "kube-proxy"
+	kubeProxyServiceAccountName                          = "open-feature-operator-kube-proxy"
+	kubeProxyServiceName                                 = "kube-proxy-svc"
+	kubeProxyImage                                       = "ghcr.io/open-feature/kube-flagd-proxy"
+	kubeProxyTag                                         = "v0.1.2"
```
Kavindu-Dodan, line 50: minor - need to bind this with renovate

webhooks/pod_webhook.go
Kavindu-Dodan: @james-milligan do you plan to introduce this through helm [1] ? 

I think this is an essential configuration - to enable/disable debug logs for flagd sidecar & kube-proxy 

[1] - https://github.com/open-feature/open-feature-operator/blob/main/chart/open-feature-operator/values.yaml#L4

Kavindu-Dodan:
Cool work. I like the approach you took with `provider: kube-proxy` option.

Look forward to the refined PR

webhooks/pod_webhook.go
```diff
@@ -230,6 +257,146 @@ func (m *PodMutator) injectSidecar(
...
+}
+
+func (m *PodMutator) handleKubeProxy(ctx context.Context, sidecar *corev1.Container, source v1alpha1.Source) error {
+	// does the proxy exist
+	exists, ready, err := m.isKubeProxyReady(ctx)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		//create, defer pod deployment
```
Kavindu-Dodan, line 369: For my understanding - what will happen if proxy creation continue to fail (ex:- image pull error, OOMKill) 🤔

webhooks/pod_webhook.go
james-milligan: yes, all configurations will follow the existing flow, using environment variable overrides for the default values, and then relevant configurations being overwritten though custom resources

IMO I don't believe there are any configurations in this implementation that require configurations via custom resources, all can be handled through the helm chart / env vars

webhooks/pod_webhook.go
james-milligan: The annotated pod requiring a sidecar will not be injected, IMO this is acceptable, as if the app was deployed without the sidecar, or without the `kube-proxy` flags, its configuration may be wrong.

james-milligan: The message `"kube-flagd-proxy is not ready, deferring pod admission"` is passed out via the deny so will be present on a describe of the replicaset, however I will add also add in an additional check to confirm whether the kube proxy deployment is not yet ready because it is in an error state and pass this out as an actual error.

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -47,12 +47,13 @@ const (
 	defaultSocketPath                string           = ""
 	defaultEvaluator                 string           = "json"
 	defaultImage                     string           = "ghcr.io/open-feature/flagd"
-	defaultTag                       string           = "v0.4.4"
+	defaultTag                       string           = "v0.4.5"
```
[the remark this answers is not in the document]
james-milligan: it looks like this was broken once we introduced multiple releases, as its looking for GitHub release tag, which now contains flagd / core / kube-flagd-proxy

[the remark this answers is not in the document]
james-milligan: https://github.com/open-feature/open-feature-operator/issues/415

chart/open-feature-operator/values.yaml
```diff
@@ -8,13 +8,23 @@ sidecarConfiguration:
...
   probesEnabled: true
+  debugLogging: false
+
+kubeProxyConfiguration:
+  port: 8015
+  metricsPort: 8016
+  image:
+     # these fields must remain in the same order, renovate uses a regex to update the tag value
+    repository: "ghcr.io/open-feature/kube-flagd-proxy"
+    tag: v0.1.2
```
james-milligan, line 26: add to renovate following https://github.com/open-feature/open-feature-operator/issues/415

webhooks/pod_webhook.go
james-milligan: ive introduced this configuration, along with port, metrics port and image/tag

james-milligan: > I think this is an essential configuration - to enable/disable debug logs for flagd sidecar & kube-proxy

adding this configuration to the sidecar requires changes to the FalgSourceConfiguration CRD, which is outside of the scope of this PR, ive opened an issue here: https://github.com/open-feature/open-feature-operator/issues/416

webhooks/pod_webhook.go
```diff
@@ -39,11 +44,24 @@ const (
...
 	ProbeInitialDelay                                    = 5
+	kubeProxyDeploymentName                              = "kube-proxy"
+	kubeProxyServiceAccountName                          = "open-feature-operator-kube-proxy"
+	kubeProxyServiceName                                 = "kube-proxy-svc"
+)
+
+var (
+	currentNamespace      = "open-feature-operator-system"
+	kubeProxyImage        = "ghcr.io/open-feature/kube-flagd-proxy"
+	kubeProxyTag          = "v0.1.2"
```
james-milligan, line 55: following https://github.com/open-feature/open-feature-operator/issues/415 this version should also be added to the renovate config

config/rbac/kustomization.yaml
toddbaert: Ya and after a release we can do some smoke testing as well.

webhooks/pod_webhook.go
Kavindu-Dodan: Perfect thanks :)

toddbaert:
I'm going to review this an manually test it today.

webhooks/pod_webhook.go
Kavindu-Dodan: This is better and support fail fast approach. This will help administrators to deal with root causes :)

webhooks/pod_webhook.go
```diff
@@ -244,6 +248,63 @@ func (m *PodMutator) injectSidecar(
 	return json.Marshal(pod)
 }
 
+func (m *PodMutator) isKubeProxyReady(ctx context.Context) (bool, bool, error) {
+	m.Client.Scheme()
```
Kavindu-Dodan, line 252: Seems to be an unused call

controllers/core/flagsourceconfiguration/controller.go
```diff
@@ -35,17 +40,50 @@ import (
 	corev1alpha1 "github.com/open-feature/open-feature-operator/apis/core/v1alpha1"
 )
 
+const (
+	OpenFeatureAnnotationPath         = "spec.template.metadata.annotations.openfeature.dev/openfeature.dev"
+	FlagSourceConfigurationAnnotation = "flagsourceconfiguration"
+	OpenFeatureAnnotationRoot         = "openfeature.dev"
```
Kavindu-Dodan, line 46: unused constants. Seems to go unnoticed by the linter 🤔

Kavindu-Dodan:
General observation. There's mixed usage of `kube-flagd-proxy`, `kube-proxy` & `flagd-kube-proxy`

If possible shall we try to agree and unify the naming? :)

toddbaert:
In my manual testing I found an issue where the `kube-proxy` source errors without a fully qualified CRD name (`{namespace}/{name}`).

For the `kube-proxy`, this works:

```yaml
kind: FlagSourceConfiguration
metadata:
    name: end-to-end-fsc
spec:
    sources:
    - source: default/end-to-end
      provider: kube-proxy
    - source: default/end-to-end2
      provider: kube-proxy
```

This does not:

```yaml
kind: FlagSourceConfiguration
metadata:
    name: end-to-end-fsc
spec:
    sources:
    - source: default/end-to-end
      provider: kube-proxy
    - source: default/end-to-end2
      provider: kube-proxy
```

The error actually comes from flagd, I think because OFO should be adding the namespace before it starts flagd:

`rpc error: code = Unknown desc = invalid resource uri format, expected <namespace>/<crdName> but got: end-to-end2`

controllers/core/flagsourceconfiguration/controller.go
```diff
@@ -104,6 +152,124 @@ func (r *FlagSourceConfigurationReconciler) Reconcile(ctx context.Context, req c
...
+			Name:      KubeProxyServiceName,
+			Namespace: r.KubeProxyConfig.Namespace,
+		},
+		Spec: corev1.ServiceSpec{
+			Selector: map[string]string{
+				"app.kubernetes.io/name": KubeProxyDeploymentName,
+			},
+			Ports: []corev1.ServicePort{
+				{
+					Name:       "flagd-kube-proxy",
```
toddbaert, line 190: Might be worth putting `flagd-kube-proxy` into a public const somewhere. We use it in some log lines elsewhere.

config/rbac/role.yaml
```diff
@@ -37,6 +37,18 @@ rules:
...
+  resources:
+  - services
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
toddbaert, line 51: Are all of these really needed?

webhooks/pod_webhook.go
```diff
@@ -244,6 +248,63 @@ func (m *PodMutator) injectSidecar(
...
+		return err
+	}
+	if !exists || (exists && !ready) {
+		return &kubeProxyDeferError{}
+	}
+
+	config := []types.SourceConfig{
+		{
+			Provider: "grpc",
+			Selector: fmt.Sprintf("core.openfeature.dev/%s", source.Source),
```
toddbaert, line 290: I _think_ this needs to be changed to prefix the namespace of the source, if not present. We may want a util function for this since I think the k8s sync is doing something similar.

See: https://github.com/open-feature/open-feature-operator/pull/412#issuecomment-1487439625

toddbaert:
I think a few changes are needed but this is REALLY cool, IMHO. I love how the proxy starts dynamically.

controllers/core/flagsourceconfiguration/controller.go
```diff
@@ -104,6 +152,124 @@ func (r *FlagSourceConfigurationReconciler) Reconcile(ctx context.Context, req c
...
+func (r *FlagSourceConfigurationReconciler) handleKubeProxy(ctx context.Context) error {
+	exists, err := r.doesKubeProxyExist(ctx)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		return r.deployKubeProxy(ctx)
+	}
+	return nil
+}
```
toddbaert, line 164: cool

pkg/types/source-config.go
```diff
@@ -0,0 +1,14 @@
+package types
+
+// SourceConfig is a 1:1 mapping for flagd SourceConfig type. JSON encoded SourceConfig becomes a startup argument for
+// flagd
+// NOTE - once we have API stability, make this a dependency at flagd to avoid duplication
+type SourceConfig struct {
```
Kavindu-Dodan, line 6: Minor - needs to bump this to match the latest changes [1] :) 

[1] - https://github.com/open-feature/flagd/blob/main/core/pkg/runtime/from_config.go#L51

docs/flagd_kube_proxy.md
```diff
@@ -0,0 +1,53 @@
+# flagd-kube-proxy
```
AlexsJones, line 1: Calling this kube-proxy is so confusing as kube-proxy is a kubernetes component 

https://kubernetes.io/docs/reference/command-line-tools-reference/kube-proxy/

Kavindu-Dodan: Given that what we really do here is to proxy CRD access, another alternative is to call this `flagd-crd-proxy` 🤔

james-milligan: @AlexsJones are you happy with `flagd-crd-proxy`, I can make these changes tomorrow

james-milligan:
> General observation. There's mixed usage of `kube-flagd-proxy`, `kube-proxy` & `flagd-kube-proxy`
> 
> If possible shall we try to agree and unify the naming? :)

ive consolidated all references to be `kube-flagd-proxy` matching the name in the flagd repository.

If we wish to change the name, as per @AlexsJones comment, I suggest we do so in flagd first, and before the next release

config/rbac/flagd_kubernetes_sync_clusterrolebinding.yaml
```diff
@@ -7,6 +7,10 @@ subjects:
   name: open-feature-operator-controller-manager
   namespace: system
   apiGroup: ""
+- kind: ServiceAccount
+  name: kube-proxy
+  namespace: system
```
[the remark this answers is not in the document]
james-milligan: Ive renamed this SA to `open-feature-operator-kube-flagd-proxy`, however the kube-flagd-proxy is deployed to the same namespace as the operator, as such `namespace: system` is suffice

controllers/core/flagsourceconfiguration/controller.go
```diff
@@ -35,17 +40,50 @@ import (
...
 type FlagSourceConfigurationReconciler struct {
 	client.Client
 	Scheme *runtime.Scheme
 	// ReqLogger contains the Logger of this controller
-	Log logr.Logger
+	Log             logr.Logger
+	KubeProxyConfig *KubeProxyConfiguration
+}
+
+type KubeProxyConfiguration struct {
```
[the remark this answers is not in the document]
james-milligan: nice suggestion, will implement this once I have handled the other comments

webhooks/pod_webhook.go
james-milligan: resolved

docs/flagd_kube_proxy.md
beeme1mr: What do you think about calling this `flagd-proxy`? While it's intended for Kubernetes, it's not a hard requirement. That should address the concern raised by Alex and provide flexibility if the proxy is required for other deployment patterns.

james-milligan: I strongly agree with switching to `flagd-proxy`, it decouples us nicely from any underlying flag sources

toddbaert: I'm OK with `flagd-proxy`.

Kavindu-Dodan (approved):

controllers/core/flagsourceconfiguration/kube-flagd-proxy.go
```diff
@@ -0,0 +1,205 @@
...
+		kph.config.Tag = kpt
+	}
+	portString, ok := os.LookupEnv(envVarProxyPort)
+	if ok {
+		port, err := strconv.Atoi(portString)
+		if err != nil {
+			return kph, fmt.Errorf("could not parse %s env var: %w", envVarProxyPort, err)
+		}
+		kph.config.Port = port
+	}
```
thisthat, line 65: since this is duplicated it could be a handy function in the util pkg
something like `getIntEnv(envVar string, default int) (int, error)`

thisthat (approved):
LGTM ✅

toddbaert (approved):

PR #414, opened 2023-03-23
chore: improve container build layer caching

mowies:
### This PR
- optimizes container builds by moving the reading of build arguments to right before they are needed
- with the next `RUN` command after an `ARG` command, the docker layer cache is always invalidated, since the RUN command can have different effects with different build args that the image builder cannot foresee.
- by moving the build arg steps further down in the dockerfile, caching should be improved by a lot
- I also enabled github actions caching for the docker builds to improve pipeline times


james-milligan (approved):
🫡

toddbaert:
This is a very nice catch. Thanks @mowies

PR #420, opened 2023-03-27
chore: introduce additional unit tests for api packages

### LINKED ISSUE #394: Create Unit tests for Controllers using K8s Fake Client
# Short abstract

The goal of this issue is to add unit tests for the `FeatureFlagConfiguration` and `FlagSourceConfiguration` controllers, using the Kubernetes [fake client](https://pkg.go.dev/k8s.io/client-go/kubernetes/fake).

# Why 

Adding these tests will help to better verify the behavior of the controllers, as opposed to the ginkgo component tests, since with this approach it's possible to check for certain assertions after each invocation of the controller's `Reconcile` method, and controlling the current state of the relevant resources with the fake kubernetes client.

# Functional requirements

- [ ] Create unit tests using the K8s fake client for both controllers

odubajDT:
## This PR

- refactor api packages
- cover api packages by unit tests
- adapt changes

### Related Issues

Fixes #394


apis/core/v1alpha1/featureflagconfiguration_types.go
```diff
@@ -110,7 +108,7 @@ func init() {
 	SchemeBuilder.Register(&FeatureFlagConfiguration{}, &FeatureFlagConfigurationList{})
 }
 
-func GetFfReference(ff *FeatureFlagConfiguration) metav1.OwnerReference {
+func (ff *FeatureFlagConfiguration) GetReference() metav1.OwnerReference {
```
toddbaert, line 111: I guess this was just a style/consistency choice since we added receivers to other functions here?

I'm still relatively new to go. If there was more of a reason for this, let me know!

toddbaert (approved):
lgtm

apis/core/v1alpha1/featureflagconfiguration_types.go
odubajDT: Basically yes, only a style/consistency adaptation, but I was also referring to SOLID principles here. Basically the FeatureFlagConfiguration object has all the information needed to get the owner references. Therefore I found it as a good idea to define a simple getter to retrieve this information from the object.

But generally yes, in this case it's really a choice of styling/consistency and does not have any behavioral effect on the code.

thisthat: There is also another benefit to this change. Before, a pointer to a FFConf struct was expected as input. This allows calling the function with a nil argument triggering an invalid memory access. Now, you need to have a valid FFConf struct before you can call the function

thisthat (approved):

PR #434, opened 2023-03-31
feat: introduce debugLogging parameter to FlagSourceConfiguration CRD

### LINKED ISSUE #416: Add debugLogging field to flagSourceConfiguration
Add a bool field to the FlagSourceConfguration CRD to control the addition of the `--debug` flag to the injected sidecar

odubajDT:
## This PR

- introduces new field to FlagSourceConfiguration
- adapts mutating webhook to inject the flags to the sidecar
- adapts conversion webhooks of FlagSourceConfiguration
- cover added/changed functionality by tests

### Related Issues

Fixes #416


beeme1mr (dismissed):
Can these values be set via Helm? Is so, can you please document this [here](https://github.com/open-feature/open-feature-operator/tree/main/chart/open-feature-operator)?

agardnerIT:
Where is this change documented? Please do not merge until it is documented.

odubajDT:
Converting back to draft, as there is a need to support setting these parameters via helm and also document the changes accordingly

odubajDT:
> Can these values be set via Helm? Is so, can you please document this [here](https://github.com/open-feature/open-feature-operator/tree/main/chart/open-feature-operator)?

If we want to set these values via helm, we have to introduce a env variable to set them, do we want this? I consider these flags more of a development options (at least the rawSidecarArgs), so maybe there only the debugLogging option? Sounds like a dev variable to me as well... But I will leave the decision on the team

beeme1mr (approved):
Looks good, thanks. Also, rour comment about Helm makes sense and I agree. 👍

webhooks/pod_webhook.go
```diff
@@ -230,6 +252,19 @@ func (m *PodMutator) injectSidecar(
 		}
 	}
 
+	// append raw sidecar container args
+	if flagSourceConfig.RawSidecarArgs != nil {
+		sidecar.Args = append(sidecar.Args, flagSourceConfig.RawSidecarArgs...)
+	}
```
thisthat, line 258: I would have raw argument either as the first thing or last thing added to the array. I suggest adding them with the least priority over existing arguments. My motivation is that if I configured a flag directly, eg `debugLogging=true`, and also via `rawSidecarArgs`, the direct value should win to avoid confusion.

toddbaert (approved):
LGTM

docs/flag_source_configuration.md
```diff
@@ -54,6 +56,8 @@ The relevant `FlagSourceConfigurations` are passed to the operator by setting th
...
 | ProbesEnabled       | Enable or disable Liveness and Readiness probes of the flagd sidecar. When enabled, HTTP probes( paths - `/readyz`, `/healthz`) are set with an initial delay of 5 seconds                  | optional `boolean`                                                        | `true`                       |       
+| DebugLogging        | Enable or disable --debug flag of flagd sidecar                                                                                                                                             | optional `boolean`                                                        | `false`                      |
+| RawSidecarArgs      | String arguments passed to the flagd sidecar on startup, flagd documentation can be found [here](https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md)       | optional `array of strings`                                               | `""`                         |
```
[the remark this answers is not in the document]
odubajDT: I tried applying multiple parameters to the flagD binary and the current behaviour is that the latest applied value wins

```
```

webhooks/pod_webhook.go
thisthat: let's continue the discussion here: https://github.com/open-feature/open-feature-operator/pull/434#discussion_r1158042848

docs/flag_source_configuration.md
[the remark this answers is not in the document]
thisthat: Let's add them as the first thing and document that any direct value configured wins over the value set via the `RawSidecarArgs` flag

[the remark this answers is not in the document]
odubajDT: Actually we probably need to have a controlling logic in the operator. Let's take a use-case, when the user sets the `.spec.debugLogging` field to false (or does not set it at all and the default value is false) and afterwards the `.spec.rawSidecarArgs` contain `--debug` flag, which is basically the flag set when `.spec.debugLogging` is set to true. In this case, we need to parse the `.spec.rawSidecarArgs` and check for inconsistent states. 

In the current state I am actually questioning my original idea of adding `.spec.rawSidecarArgs` to the FlagSourceConfiguration CRD and if it's worth the work. What about your thoughts @toddbaert @thisthat ?

docs/flag_source_configuration.md
```diff
@@ -54,6 +56,8 @@ The relevant `FlagSourceConfigurations` are passed to the operator by setting th
...
 | ProbesEnabled       | Enable or disable Liveness and Readiness probes of the flagd sidecar. When enabled, HTTP probes( paths - `/readyz`, `/healthz`) are set with an initial delay of 5 seconds                  | optional `boolean`                                                        | `true`                       |       
+| DebugLogging        | Enable or disable --debug flag of flagd sidecar                                                                                                                                             | optional `boolean`                                                        | `false`                      |
+| RawSidecarArgs      | String arguments passed to the flagd sidecar on startup, flagd documentation can be found [here](https://github.com/open-feature/flagd/blob/main/docs/configuration/configuration.md). Please be aware, flags defined by CRD parameters have bigger value and raw arguments will be ignored. | optional `array of strings`                                               | `""`                         |
```
Kavindu-Dodan, line 60: Minor - Regarding the link, shouldn't we target to startup flags document [1] ? :) 

[1] - https://github.com/open-feature/flagd/blob/main/docs/configuration/flagd_start.md

docs/flag_source_configuration.md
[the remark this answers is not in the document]
Kavindu-Dodan: IMO, adding `DebugLogging` is essential. However, given the complexity of parsing & validations, may be it is better to avoid `rawSidecarArgs ` 🤔  If we really want to control the flagd startup flags, then it is best exposed through CRD, so that we know how to control that flag.

docs/flag_source_configuration.md
odubajDT: good catch, but probably the whole `rawSidecarArgs` will be removed

docs/flag_source_configuration.md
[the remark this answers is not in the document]
odubajDT: already reverted the changes of `rawSidecarArgs` (last commit in this PR), but still waiting for input from @toddbaert

apis/core/v1alpha1/flagsourceconfiguration_types.go
```diff
@@ -122,6 +123,10 @@ type FlagSourceConfigurationSpec struct {
 	// ProbesEnabled defines whether to enable liveness and readiness probes of flagd sidecar. Default true(enabled)
 	// +optional
 	ProbesEnabled *bool `json:"probesEnabled"`
+
+	// DebugLogging defines whether to enable --debug flag of flagd sidecar. Default false(disabled)
```
thisthat, line 127: nit
```suggestion
	// DebugLogging defines whether to enable --debug flag of flagd sidecar. Default false (disabled).
```

apis/core/v1alpha3/flagsourceconfiguration_types.go
```diff
@@ -88,6 +88,10 @@ type FlagSourceConfigurationSpec struct {
 	// ProbesEnabled defines whether to enable liveness and readiness probes of flagd sidecar. Default true(enabled)
 	// +optional
 	ProbesEnabled *bool `json:"probesEnabled"`
+
+	// DebugLogging defines whether to enable --debug flag of flagd sidecar. Default false(disabled)
```
thisthat, line 92: ditto

docs/flag_source_configuration.md
```diff
@@ -54,6 +55,7 @@ The relevant `FlagSourceConfigurations` are passed to the operator by setting th
...
 | RolloutOnChange     | When set to true the operator will trigger a restart of any `Deployments` within the `FlagSourceConfiguration` reconcile loop, updating the injected sidecar with the latest configuration. | optional `boolean`                                                        | `false`                      | 
 | ProbesEnabled       | Enable or disable Liveness and Readiness probes of the flagd sidecar. When enabled, HTTP probes( paths - `/readyz`, `/healthz`) are set with an initial delay of 5 seconds                  | optional `boolean`                                                        | `true`                       |       
+| DebugLogging        | Enable or disable --debug flag of flagd sidecar                                                                                                                                             | optional `boolean`                                                        | `false`                      |
```
thisthat, line 58:
```suggestion
| DebugLogging        | Enable or disable `--debug` flag of flagd sidecar                                                                                                                                             | optional `boolean`                                                        | `false`                      |
```

webhooks/pod_webhook.go
```diff
@@ -196,20 +192,36 @@ func (m *PodMutator) injectSidecar(
...
 	if err != nil {
-		return nil, err
+		return err
 	}
+	return nil
+}
 
-	err = m.appendSources(sources, &sidecar)
-	if err != nil {
+func (m *PodMutator) InjectSidecar(
```
thisthat, line 210: Why do we export this function now?

odubajDT: probably leftover, will revert thanks!

thisthat (approved):
LGTM

Kavindu-Dodan (approved):
LGTM :)

docs/flag_source_configuration.md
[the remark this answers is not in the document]
toddbaert: Ya, this is fine with me. It was a simple nice-to-have, but seems like it's added too much complexity.

Issue #377, opened 2023-03-09
Support kube-flagd-proxy deployment in OFO


toddbaert:
We need to support the kube-flagd-proxy strategy/source in our OFO deployment, in addition to the existing `kubernetes` and `flie` sources.

This involves:

- updating the CRDs to support this source
- starting the kube-flagd-proxy
- starting flagd with this source, updating the necessary reconciliations/webhooks
- tests for the above

```[tasklist]
### Tasks
- [x] Create/update FlagSourceConfiguration CRD to support `flagd-kube-proxy` deployment (we'll need to come up with a good name for this `provider`).
- [x] Enhance pod webhook to start flagd with gRPC source pointed to K8s proxy (blocked by [this](https://github.com/open-feature/open-feature-operator/issues/389))
- [x] Enhancements to `flagsourceconfiguration_controller` (and/or elsewhere?) to deploy and manage `flagd-kube-proxy`
```

toddbaert:
@james-milligan I think this is all done? Please close if you agree.

PR #436, opened 2023-04-03
fix: restrict permissions to only access specific CRB

bacherfl:
This PR reduces the required permissions for the OFO to only be able to modify the `open-feature-operator-flagd-kubernetes-sync` ClusterRoleBinding.
Once https://github.com/open-feature/open-feature-operator/issues/377 has been implemented, this permission can likely be dropped completely


thisthat (approved):
LGTM but I would double check with the helm package step if we lock down the name there as well

james-milligan (approved):

PR #425, opened 2023-03-28
feat: introduce Flagd CRD with reconciliation

skyerus:
## POC

This is a WIP implementation of a derivative of [the FlagService OFEP](https://github.com/open-feature/ofep/blob/main/OFEP-ofo-flag-service.md).

This PR introduces a CRD named `Flagd`. The creation of a `Flagd` CR triggers its reconciliation loop which does the following:
- Creates a Deployment of flagd with labels as specified by the named Service selector

## How to run the POC

The demo aims to exemplify a legitimate end-to-end deployment scenario by using [Kubernetes Gateway API (KGA)](https://gateway-api.sigs.k8s.io/) to ingest traffic to a demo application & the created flagd Deployment.

Prerequisites:

- [Kind](https://kind.sigs.k8s.io/)
- Checkout OFO at this branch
- [istioctl](https://istio.io/latest/docs/setup/getting-started/#download) (this can be substituted for any other KGA implementor but make sure to replace references to istio with the substitute)

Note: To get this to work locally I had to add hostnames to my hosts file to route them to localhost. Without this the browser doesn't send the appropriate Host header to the Gateway, preventing it from being able to route the request.

```
# build OFO
IMG=ghcr.io/open-feature/open-feature-operator:local ARCH=arm64 make docker-build
# create kind cluster
kind create cluster --config ./test/e2e/kind-cluster.yml
# load OFO into cluster
kind load docker-image ghcr.io/open-feature/open-feature-operator:local
# deploy OFO
IMG=ghcr.io/open-feature/open-feature-operator:local make deploy-operator
# install KGA CRDs
kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v0.6.2/standard-install.yaml
# install istio gateway
istioctl install --set profile=minimal -y
# deploy resources
kubectl apply -f config/samples/dev.yaml
# add hostnames to hosts file (location of hosts file is macOS specific in this example)
sudo sh -c 'echo "127.0.0.1 ofoclientapp.com" >> /private/etc/hosts'
sudo sh -c 'echo "127.0.0.1 flagd.ofoclientapp.com" >> /private/etc/hosts'
# port forward 8080 to the gateway pod
kubectl port-forward $(kubectl get pods --selector=istio.io/gateway-name=clientsidegateway -o \
  jsonpath='{.items[*].metadata.name}') 8080:80
```

Go to http://ofoclientapp.com:8080/ in your browser and click the Update color button to verify that the traffic is routed to flagd correctly.

[...]

PR #458, opened 2023-04-12
feat: reconciler for standalone FlagD CRD

bacherfl:
This PR is a PoC for the  [OFO Flag Service OFEP](https://github.com/open-feature/ofep/blob/main/OFEP-ofo-flag-service.md)

The implementation is based on the draft PR #425 - I decided to create a new PR because there have been several refactorings in the meantime, and there were too many conflicts to confidently merge the latest status of the main branch into this PR.

The main difference to the original PoC, #425 is that I removed the `DeploymentSpec` from the `Flagd CRD`, as in my opinion we should limit the customisability of that CR to some selected fields (e.g. imagePullPolicy, nodeAffinity, ...), as opposed to the complete Deployment spec where most fields will be replaced with what the Flagd controller will generate when creating the deployment of the Flagd instance.

Currently, this PR contains quite a lot of duplication, due to the fact that the injection of the flagd container into a PodSpec has previously been done in the pod mutating webhook, but is now also done in the new Flagd controller.
For the Flagd controller, I have extracted the injection of the flagd container into a PodSpec into a seperate component (`FlagdContainerInjector`), which can potentially be used in the webhook as well. In fact, I would recommend to do this refactoring as a separate PR first, before doing the proper implementation PR of the Flagd controller - otherwise the PR will become simply too big to review properly in my opinion

An example Flagd CR can be found in `config/samples/crds/flagd`. This manifest contains a `FeatureFlagConfiguration`, `FlagSourceConfiguration`,  `Service` and a `Flagd` - Applying this manifest will then result in a `Deployment` running flagd, which will be set up to use the feature flags defined in the `FeatureFlagConfiguration`/`FlagSourceConfiguration`. The deployment will be accessible via the `open-feature-dev-flagd` Service. To quickly verify this, the following commands can be used after applying the manifest:

```
kubectl port-forward -n open-feature-operator-system svc/open-feature-dev-flagd 8080:80 &
curl -X POST "localhost:8080/schema.v1.Service/ResolveString" -d '{"flagKey":"hex-color","context":{}}' -H "Content-Type: application/json"
```


beeme1mr:
@bacherfl this PR looks stale. Can it be closed?

PR #496, opened 2023-07-05
attempt to improve documentation

Kavindu-Dodan:
## This PR

Attempts to improve OFO documentation


agardnerIT (changes requested):
Thanks for these changes. They largely look great! A few comments though:

1. `docs/getting_started.md` and `docs/quick_start.md` should be combined. They serve largely the same purpose and as such, having both is confusing and redundant.
2. At the end of `docs/installation.md` we should link to `docs/getting_started.md` otherwise, after installation, I don't know what else to do.
3. At the end of `docs/quick_start.md` (or wherever that content ends up, we should mention that the OFO uses flagd as a flag provider and thus all the flagd documentation is also relevant. One specific callout is to point of the fact that `ResolveString` isn't the only option, but [other options are available](https://github.com/open-feature/flagd/blob/main/docs/usage/evaluation_examples.md).

Kavindu-Dodan:
> Thanks for these changes. They largely look great! A few comments though:
> 
> 1. `docs/getting_started.md` and `docs/quick_start.md` should be combined. They serve largely the same purpose and as such, having both is confusing and redundant.
> 2. At the end of `docs/installation.md` we should link to `docs/getting_started.md` otherwise, after installation, I don't know what else to do.
> 3. At the end of `docs/quick_start.md` (or wherever that content ends up, we should mention that the OFO uses flagd as a flag provider and thus all the flagd documentation is also relevant. One specific callout is to point of the fact that `ResolveString` isn't the only option, but [other options are available](https://github.com/open-feature/flagd/blob/main/docs/usage/evaluation_examples.md).

thank you @agardnerIT for the feedback. I addressed all your points 

1. Kept one quick start guide : this is now comprehensive and includes minimal installation instructions
2. Added the link to quick start guide : operator installation can be skipped but users can follow other steps to deploy a workload
3. Added the `what's next` section [1] linking to other resources 

[1] - https://github.com/open-feature/open-feature-operator/pull/496/files#diff-61f4196322cfa9f52c5f548f13b7f781243b2cae898862a8688462d914ba7173R178-R182

docs/README.md
```diff
@@ -1,24 +1,41 @@
...
+Follow the detailed installation guide to deploy open feature operator to your local cluster, 
 
 - [Installation](./installation.md)
-- [Getting Started](./getting_started.md)
 
 ## Configuration
 
-Configuration of the deployed sidecars is handled through the `FeatureFlagConfiguration` CRs defined in the `openfeature.dev/featureflagconfiguration` annotation of a deployed `PodSpec`. 
-> Further configuration of the operator will be possible in the future, to help contribute [click here](https://github.com/open-feature/open-feature-operator/issues)
+Configuration of the deployed sidecars is handled through the `FeatureFlagConfiguration` custom resources defined in the `openfeature.dev/featureflagconfiguration` annotation of a deployed `PodSpec`.
```
toddbaert, line 15: I think this isn't really the best explanation.

`FeatureFlagConfiguration` actually is inteded most to define flags, while `FlagSourceConfiguration` controls the sidecars.

docs/quick_start.md
```diff
@@ -0,0 +1,182 @@
+## Quick Start for OpenFeature Operator
```
toddbaert, line 1: I think this is a really nice addition. Using busy-box, having no ingress, etc, is nice b/c it's very "barebones".

toddbaert (changes requested):
This is an improvement, but I think there's a mistake [here](https://github.com/open-feature/open-feature-operator/pull/496/files#r1255857908).

docs/README.md
Kavindu-Dodan: Thank you 🙌 

A good catch and I think naming is a little misleading and confusing sometimes. 

I updated the documentation to correct this

docs/quick_start.md
Kavindu-Dodan: Agree, we should keep it minimal and let users get things up and running fast. Later, we can refer them to detailed guides

Kavindu-Dodan:
> This is an improvement, but I think there's a mistake [here](https://github.com/open-feature/open-feature-operator/pull/496/files#r1255857908).

Thank you. I corrected them and also expanded `Flag Source configuration` section [1] with other flag sources and configuration options. Strangely, we never documented all the options we supported  

[1] - https://github.com/open-feature/open-feature-operator/blob/b7a6b531ed9ab6edfe5ecf6038bfdc2ae570a811/docs/flag_source_configuration.md

docs/README.md
```diff
@@ -1,24 +1,41 @@
-# Docs
+# Documentation
 
-This directory contains all OpenFeature Operator documentation, see table of contents below:
+This directory contain OpenFeature Operator documentation.
```
thisthat, line 3:
```suggestion
This directory contains OpenFeature Operator documentation. 
```

docs/flag_source_configuration.md
```diff
@@ -1,7 +1,96 @@
...
+  port: 8080                      # port of the flagd sidecar
+```
+
+## Feature flag sources
+
+This section explains how to configure feature flag sources to injected flag sidecar.
+
+`FlagSourceConfiguration` support multiple flag sources. Sources are configured as a list and given below are supported sources and their configurations,
+
+### kubernetes aka `FeatureFlagConfiguration`
```
toddbaert, line 26: both this, and `flagd-proxy` make use of `FeatureFlagConfiguration`. I think you might want to put these two closer together, and conceptually link and contrast them (probably just highlight that they both use `FeatureFlagConfiguration`, but one uses a proxy and doesn't require direct access of the workload to k8s).

toddbaert (approved):
LGTM, but I'd like to see this last small change: https://github.com/open-feature/open-feature-operator/pull/496/commits/1dbf3d0745bf0e33e6090297c4a46beb34b84389#r1258239251

docs/flag_source_configuration.md
Kavindu-Dodan: Good point. Added a description and reordered the source list to keep k8s & proxy approach close to each other

toddbaert:
@Kavindu-Dodan the last set of changes looks great! I think after it's merged we could put @thisthat 's [changes](https://github.com/open-feature/open-feature-operator/pull/497) in/around these changes as a sub-point.

toddbaert (approved):
@Kavindu-Dodan the last set of changes looks great! I think after it's merged we could put @thisthat 's [changes](https://github.com/open-feature/open-feature-operator/pull/497) in/around these changes as a sub-point.

docs/annotations.md
```diff
@@ -25,6 +31,10 @@ Example:
       openfeature.dev/flagsourceconfiguration: "config-A, config-B"
 ```
 
+## Deprecated annotations
+
+Given below are references to **deprecated** annotations used by previous versions of the operator.
+
 ### `openfeature.dev/allowkubernetessync`
```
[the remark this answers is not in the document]
Kavindu-Dodan: Good catch, this is internal to the operator and worth documenting.

docs/annotations.md
```diff
@@ -26,9 +32,14 @@ Example:
 ```
 
 ### `openfeature.dev/allowkubernetessync`
-*This annotation is used internally by the operator.*  
-This annotation is used to mark pods which should have their permissions backfilled in the event of an upgrade. When the OFO manager pod is started, all `Service Accounts` of any `Pods` with this annotation set to `"true"` will be added to the `flagd-kubernetes-sync` `Cluster Role Binding`.
+*This annotation is used INTERNALLY by the operator.*
```
beeme1mr, line 35:
```suggestion
*This annotation is used INTERNALLY by the operator.*
```

beeme1mr (approved):

Issue #517, opened 2023-08-07
Consider restricting FlagSourceConfiguration CRD


Kavindu-Dodan:
Pre-requisites:

- [x] https://github.com/open-feature/open-feature-operator/issues/529
- [x] https://github.com/open-feature/flagd/issues/983
- [x] https://github.com/open-feature/open-feature-operator/issues/531

This issue was derived from comment https://github.com/open-feature/open-feature-operator/pull/514#issuecomment-1665877515 

OpenFeature operator supports configurability of flagd sidecar image & tag through two options,

- As startup arguments of the operator (using helm/K8s yaml) [^1]
- Using FlagSourceConfiguration's image & tag override options [^2]

The latter has a potential security impact as the side-car injection can be controlled based on the CRD definition of the respective deployment. 

**Solution**

For a given operator deployment, it is best to fix the sidecar image & tag and disallow any other overloading options. This means, keeping the Operator startup option and removing the option provided through `FlagSourceConfiguration` crd.

**Discussion**

Consider following when doing this,

- This is best done with beta version of the Operator API https://github.com/open-feature/open-feature-operator/issues/352 
- Consider migration path, warnings and test implementation impact 

[^1]: https://github.com/open-feature/open-feature-operator/blob/v0.2.35/main.go?rgh-link-date=2023-08-03T18%3A11%3A45Z#L93-L109
[^2]: https://github.com/open-feature/open-feature-operator/blob/v0.2.35/docs/flag_source_configuration.md#sidecar-configurations

odubajDT:
The configurability of the flad sidecar via startup arguments (image & tag) is not part of the implementation yet and needs to be implemented separately. I would probably consider implementing this properly with tests after the implementation of v1beta1 is merged (but definitely before it's released, so we do not have breaking changes). WDYT @Kavindu-Dodan @toddbaert ?

Kavindu-Dodan:
@odubajDT Actually this is already there but hidden inside a configuration loader [^1] and not well document 

Going forward we should only allow this configuration loading (+ document this behaviour) and disallow change through CRD 


[^1]: - https://github.com/open-feature/open-feature-operator/blob/v0.2.35/controllers/common/flagd-proxy.go#L68-L75

odubajDT:
> @odubajDT Actually this is already there but hidden inside a configuration loader [1](#user-content-fn-1-51f87aad3da9155b04a47936f7360f75) and not well document
> 
[...]

Issue #433, opened 2023-03-30
Discrepancy of OFO & flagd on file source naming  


Kavindu-Dodan:
Pre-requisites:

- [x] https://github.com/open-feature/open-feature-operator/issues/529
- [ ] https://github.com/open-feature/flagd/issues/983
- [ ] https://github.com/open-feature/open-feature-operator/issues/531

**Background**

flagd is capable of accepting file-based feature flag configurations. 

From flagd documentation, file-based configurations are accepted with URI or Source with `file` prefix [1] .

`flagd start --uri file:etc/featureflags.json` 

or sources yaml [2]
```
- uri: config/samples/example_flags.json
    provider: file
```

 In OFO however, `FlagSourceConfiguration` require the provider to be `filepath` [3]

```
  - source: /etc/flags.json
    provider: filepath
```

This mixed usage has the potential for confusion. 

**Improvement**

We should aim to unify this terminology, agreeing to a single name (similar to `http`, `grpc` and `kubernets`). Given that this is a breaking change, we could make this a part of beta version of the API with backward compatiblity. 

[1] - https://flagd.dev/reference/sync-configuration/
[2] - https://flagd.dev/reference/sync-configuration/#source-configuration
[3] - https://github.com/open-feature/open-feature-operator/blob/main/docs/flag_source_configuration.md#source-fields

odubajDT:
This issue was resolved by https://github.com/open-feature/open-feature-operator/pull/535 in [this](https://github.com/open-feature/open-feature-operator/pull/535/files#diff-2be71624b02bf8c81b8480b76adb6eb84f949ef9307614a5c71a8939643c04e5R7) place

Issue #352, opened 2023-02-15
[research] Consider moving CRDs to beta, and making beta "hub" versions


toddbaert:
In an effort move to 1.0, we should consider moving our CRDs to beta to imply stability. Things we should consider:

- releasing beta CRD versions (all CRDs)
- making the beta version the conversion "hub" so those objects are referenced in most functional code
- making sure documentation is clear
- deprecating all other CRDs
- removing deprecated APIs

toddbaert:
FYI @odubajDT, this sounds like something you might be interested in.

odubajDT:
I would love to do this, but I guess the team needs to decide on the specs of the resources for beta

toddbaert:
> I would love to do this, but I guess the team needs to decide on the specs of the resources for beta

Yes, it's not a "right now" issue. I just wanted to you to aware of it. I think once we complete (or get closer to completing) the flagd-kube-proxy, we will have a better candidate for the beta CRD.

odubajDT:
Great, @toddbaert would it be possible to organize a team meeting to align on the specification of the CRDs ? Therefore I can start with this ticket and create a PoC

Kavindu-Dodan:
My take on this is to wait till we finalize kube-proxy related implementations. We will see more changes with proxy deployment and in my view, it's best to push those changes to alpha version. 

Once we have the proxy changes and stability on the solution, we can draft the beta version of the API. 

Timeline wise, this is 1 ~ 2 sprints of waiting

Kavindu-Dodan:
Consider addressing https://github.com/open-feature/open-feature-operator/issues/433 along with this change

thisthat:
Thanks all for the effort 🙌  I have some questions about the points listed by @odubajDT 

> v1beta1 (and therefore OFO 1.x)

why a stable release with a beta API version? I would vote for having stable APIs for OFO 1.x

> clean up leftover logic in v1alpha* API, but keep the CRD definitions

what type of clean up do you envision? 🤔 I would expect to drop any code line that deals with these CRDs

odubajDT:
> Thanks all for the effort 🙌 I have some questions about the points listed by @odubajDT
> 
> > v1beta1 (and therefore OFO 1.x)
> 
> why a stable release with a beta API version? I would vote for having stable APIs for OFO 1.x
> 
True, but we actually cannot skip `beta` API version, maybe worth waiting for some time before releasing OFO 1.x with a stable API? @toddbaert 
[...]

PR #526, opened 2023-10-18
feat: introduce v1beta1 version

odubajDT:
## This PR

- adds v1beta1 API

### Related Issues

Part-of #352 
Part-of #433 

### Follow-up
- use v1beta1 in the controllers/webhooks
- #517 
- #433 

### Notes
There is a problem with lint CI check, which is present on the main branch. This will be handled in a separate PR


odubajDT:
There is a problem with lint CI check, which is present on the main branch. This will be handled in a separate PR

apis/core/v1beta1/flagsourceconfiguration_types.go
```diff
@@ -0,0 +1,184 @@
...
+	// +optional
+	Evaluator string `json:"evaluator"`
+
+	// Image allows for the sidecar image to be overridden, defaults to 'ghcr.io/open-feature/flagd'
+	// +optional
+	Image string `json:"image"`
+
+	// Tag to be appended to the sidecar image, defaults to 'main'
+	// +optional
+	Tag string `json:"tag"`
```
Kavindu-Dodan, line 58: This is related to issue https://github.com/open-feature/open-feature-operator/issues/517 

This can be considered as a security concern and hence we can discuss on removing these 

cc - @toddbaert  @thisthat appreciate your opinion

apis/core/v1beta1/flagsourceconfiguration_types.go
```diff
@@ -0,0 +1,184 @@
...
+package v1beta1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	SyncProviderKubernetes SyncProviderType = "kubernetes"
+	SyncProviderFilepath   SyncProviderType = "filepath"
```
Kavindu-Dodan, line 26: Please consider the diff highlighted in this related issue - https://github.com/open-feature/open-feature-operator/issues/433

odubajDT: Do we already have a unified solution? I see that the ticket does not clearly state which option to use

apis/core/v1beta1/flagsourceconfiguration_types.go
odubajDT: adapted the code to remove tag and image from `FlagSourceConfiguration` CRD

toddbaert: Ya I'm fine with removing them. I have only heard agreements on this.

apis/core/v1beta1/flagsourceconfiguration_types.go
toddbaert: @beeme1mr @Kavindu-Dodan @thisthat which is better here? Any opinions? I suppose since we're already doing breaking changes here, it makes sense to concentrate them all here... so perhaps we should conform here in OFO to the flagd style: `file`.

apis/core/v1beta1/flagsourceconfiguration_types.go
odubajDT: With further implementation I came across some issues regarding this. Read comment here https://github.com/open-feature/open-feature-operator/issues/517#issuecomment-1770941958

I would suggest that covering this should be part of a separate PR, as it will bump the complexity of the PR, which will be adapting the code to use `v1beta1` instead of `v1alpha1`. Any objections on this ?

apis/core/v1beta1/flagsourceconfiguration_types.go
odubajDT: I would rather not put everything into one PR, it will make it hard to review and also hard to find the changes when we will need to locate it in the future. I would go for a separate PR, which will be a follow-up of this one. In fact one PR should not cover more that one issue.

odubajDT: adapted to `file`

apis/core/v1beta1/flagsourceconfiguration_types.go
toddbaert: I agree we can adapt to use `v1beta1` in a new PR, once we have that. OFO is locked to a particular version of flagd, so we shouldn't have breakage there.

It will be a breaking change in flagd though, so we should note that (with a `feat!`) so things are versioned accordingly.

apis/core/v1beta1/flagsourceconfiguration_types.go
toddbaert: cc @Kavindu-Dodan @beeme1mr

odubajDT:
Closing in favor of https://github.com/open-feature/open-feature-operator/pull/535

PR #535, opened 2023-10-25
feat: Introduce v1beta1 API version

### LINKED ISSUE #529: Introduce `v1beta1` API version
## Goal

The goal of this issue is to introduce the new API version `v1beta1`.

### Details

The `v1beta1` version should have 2 CRDs:

- `FeatureFlag` - the definition should be the same as `v1alpha2.FeatureFlagConfiguration`; reference [here](https://github.com/open-feature/open-feature-operator/pull/527/files#diff-1357321c0c02bc78a87dd757144a9a74b98a8de7d5f6274e1710bd4cb0385873)
- `SidecarConfiguration` - the definition should be the same as `v1alpha3.FlagSourceConfiguration`; reference [here](https://github.com/open-feature/open-feature-operator/pull/527/files#diff-e366a1e43c67559d085ce8878ecdc58fe14d78bdcc4114086a6bd6541e192364)

### Acceptance Criteria

- [x] new version `v1beta1` contains `FeatureFlag` and `SidecarConfiguration` CRDs
- [x] the new CRDs can be applied to the cluster where OFO is installed

### DoD

- [x] added `v1beta1` version with the mentioned CRDs
- [x] E2E test that the CRDs can be applied to the cluster

### Notes

To add the CRDs into the new version, use [kubebuilder](https://book.kubebuilder.io/introduction)

odubajDT:
## This PR

- adds v1beta1 API version with new resources
- adds E2E tests to check that new resources can be applied to cluster

### Related Issues

Fixes #529


PROJECT
```diff
@@ -43,4 +43,20 @@ resources:
   kind: FlagSourceConfiguration
   path: github.com/open-feature/open-feature-operator/apis/core/v1alpha3
   version: v1alpha3
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: openfeature.dev
+  group: core
+  kind: FeatureFlag
```
toddbaert, line 51: My only hesitation of this is the pluralization. It makes it sound like a single flag, which of course it isn't.

Is there a precedent for a `kind` with a plural name: ie `FeatureFlags`?

config/crd/bases/core.openfeature.dev_featureflags.yaml
```diff
@@ -0,0 +1,81 @@
...
+  creationTimestamp: null
+  name: featureflags.core.openfeature.dev
+spec:
+  group: core.openfeature.dev
+  names:
+    kind: FeatureFlag
+    listKind: FeatureFlagList
+    plural: featureflags
+    shortNames:
+    - ffc
```
toddbaert, line 16:
```suggestion
    - ff
```
I don't think ffc makes sense anymore with our new name.

PROJECT
Kavindu-Dodan: There's a discussion going on here [1] regarding the naming and I think we can't do much here as `kind` requires the singular name (see names section explanation [2])

Another alternative is to use a suffix like `cfg` - `FeatureFlagCfg` but that's up to discussion. 

[1] - https://github.com/open-feature/open-feature-operator/issues/352
[2] - https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition

PROJECT
```diff
@@ -43,4 +43,20 @@ resources:
...
+  path: github.com/open-feature/open-feature-operator/apis/core/v1beta1
+  version: v1beta1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: openfeature.dev
+  group: core
+  kind: SidecarConfiguration
```
Kavindu-Dodan, line 59: I am happy with this naming

apis/core/v1beta1/common.go
```diff
@@ -0,0 +1,11 @@
+package v1beta1
+
+type SyncProviderType string
+
+const (
+	SyncProviderKubernetes SyncProviderType = "kubernetes"
+	SyncProviderFilepath   SyncProviderType = "file"
```
Kavindu-Dodan, line 7: +1 for the fix from `filepath`

config/crd/bases/core.openfeature.dev_featureflags.yaml
Kavindu-Dodan: There's a todo here - https://github.com/open-feature/open-feature-operator/pull/535/files#diff-c29b273426465350caa0db6a338509afcf85b7400bccfffd4bec81465c2e8c41R60 

should be gone when alpha support is removed

apis/core/v1beta1/featureflag_types.go
```diff
@@ -0,0 +1,85 @@
+/*
```
Kavindu-Dodan, line 1: Regarding validation on flag schema 

alpha version uses a validation webhook [1] [2] and I think this is useful even if we are now handling flag configurations with `Flag` type. 

was there any reason to avoid the validation webhook ? Otherwise let's add that if there are no downsides to it. 

[1] - https://github.com/open-feature/open-feature-operator/blob/main/main.go#L281-L284
[2] - https://github.com/open-feature/open-feature-operator/blob/main/webhooks/featureflagconfiguration_webhook.go

Kavindu-Dodan:
Nice work. I left few comments but [this](https://github.com/open-feature/open-feature-operator/pull/535#discussion_r1373540017) IMO is important as we always try to validate flag configurations against the schema.

apis/core/v1beta1/featureflag_types.go
odubajDT: the validation webhook used to validate, that we have a valid json schema in the CRD, as in v1alpha1, the field was of type `string`, which means the user can basically put anything into this field. Currently we have a struct https://github.com/open-feature/open-feature-operator/pull/535/files#diff-c29b273426465350caa0db6a338509afcf85b7400bccfffd4bec81465c2e8c41R40 which already has a kubebuilder validations set (see the kubebuidlder comments). Therefore I assume this is even stricter validation than the previous one with a separate validation webhook

config/crd/bases/core.openfeature.dev_featureflags.yaml
odubajDT: Exactly, this if we change it to `ff`, we will get CRD installation problems, as 2 CRDs have the same shortcut name

apis/core/v1beta1/featureflag_types.go
Kavindu-Dodan: Thanks for the note on validation annotations. 

Regarding the usage of structure, how flexible will this be ? For example, if we include a new field or make an exisiting field optional (ex:- https://github.com/open-feature/flagd-schemas/issues/80 where there were discussions on requirement of the variant). If we define validations at operator side, we will need a new new version of the CRD. So wouldn't it be better to stick to schema validation even if that's primitive compared to stricter rules 🤔  ?

@toddbaert @thisthat appreciate discussion on this

odubajDT: We can always perform changes in `beta` versions, even breaking changes. Firstly in stable version breaking changes are not allowed.

In the API definition I went from the latest alpha version `v1alpha2.FeatureFlagConfiguration`, as that is the logical order to take here. 

To the flexibility, we can change basically anything (what I do not expect), but for example adding new fields is not considered as a breaking change and I do not see a problem here. When we perform a change (for example adding a new field), we will create a new `v1beta2` version, which will become the new hub version and potentially also create conversion webhooks (if needed). But this is a standard process in k8s. I would personally prefer a struct to a plain string in the definition

PROJECT
toddbaert: I agree with this naming then, and I'm quite happy with it now.

toddbaert:
I'm ready to approve once this is adapted as in https://github.com/open-feature/open-feature-operator/issues/352#issuecomment-1787284028

apis/core/v1beta1/featureflag_types.go
Kavindu-Dodan: @odubajDT yup, agree on all the points. I am also fine with updating the crd version when there's a substantial flag definition change. 

@toddbaert  do you see any other aspect on this ? Otherwise let's move ahead with this change.

toddbaert: No, I agree. With the proper struct our validation doesn't add anything.

Some day we might add validation for rule logic, but we don't have that yet.

PROJECT
beeme1mr: It looks like this will change to `FeatureFlagSource` based on [this thread](https://github.com/open-feature/open-feature-operator/issues/352#issuecomment-1784552206). I think that's a better name because the source information will also be valuable to the in-process version of flagd.

apis/core/v1beta1/featureflag_types.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+// TODO change to `ff` when v1alpha* is removed
+//+kubebuilder:resource:shortName="ffc"
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
+
+// FeatureFlag is the Schema for the featureflags API
+type FeatureFlag struct {
+	metav1.TypeMeta   `json:",inline"`
```
beeme1mr, line 67: Is it okay to have a `,` before inline?

apis/core/v1beta1/featureflag_types.go
```diff
@@ -0,0 +1,85 @@
...
+	// +kubebuilder:validation:Type=object
+	// Targeting is the json targeting rule
+	Targeting json.RawMessage `json:"targeting,omitempty"`
+}
+
+// FeatureFlagStatus defines the observed state of FeatureFlag
+type FeatureFlagStatus struct {
+}
+
+// TODO change to `ff` when v1alpha* is removed
```
beeme1mr, line 60: Please create an issue for this.

apis/core/v1beta1/featureflag_types.go
odubajDT: yes, this is expected

odubajDT:
@toddbaert @Kavindu-Dodan @beeme1mr PR adapted according to the discussion in https://github.com/open-feature/open-feature-operator/issues/352

apis/core/v1beta1/featureflag_types.go
odubajDT: issue created here https://github.com/open-feature/open-feature-operator/issues/537

config/crd/bases/core.openfeature.dev_featureflagsources.yaml
```diff
@@ -0,0 +1,303 @@
+---
```
toddbaert, line 1: Is it convention to pluralize the name and filename? `featureflagsources` instead of `featureflagsource`?

toddbaert (approved):
Approved, but I noticed [this](https://github.com/open-feature/open-feature-operator/pull/535/commits/b188fe6dc19d261ff21f238875f46d84a534283d#r1381656765) (just a question).

Also I'm noticing a lint error in the CI which almost make it seems like a dependency might require a higher go version or something: https://github.com/open-feature/open-feature-operator/actions/runs/6730203336/job/18292441905?pr=535#step:6:493

config/crd/bases/core.openfeature.dev_featureflagsources.yaml
odubajDT: this is a completely autogenerated file, it is expected here

odubajDT:
> Approved, but I noticed [this](https://github.com/open-feature/open-feature-operator/pull/535/commits/b188fe6dc19d261ff21f238875f46d84a534283d#r1381656765) (just a question).
> 
> Also I'm noticing a lint error in the CI which almost make it seems like a dependency might require a higher go version or something: https://github.com/open-feature/open-feature-operator/actions/runs/6730203336/job/18292441905?pr=535#step:6:493

Thank you!

The error is present also in the main branch and is not related to this PR. I would say it should be resolved in a separate PR. I can have a look at it.

Kavindu-Dodan (approved):

Kavindu-Dodan:
> > Approved, but I noticed [this](https://github.com/open-feature/open-feature-operator/pull/535/commits/b188fe6dc19d261ff21f238875f46d84a534283d#r1381656765) (just a question).
> > Also I'm noticing a lint error in the CI which almost make it seems like a dependency might require a higher go version or something: https://github.com/open-feature/open-feature-operator/actions/runs/6730203336/job/18292441905?pr=535#step:6:493
> 
> Thank you!
> 
> The error is present also in the main branch and is not related to this PR. I would say it should be resolved in a separate PR. I can have a look at it.

The error comes with the usage of golangci-lint@latest - From version [v1.54.2](https://github.com/golangci/golangci-lint/releases/tag/v1.54.2) onwards, there's a transitive dependency which require Go version 1.20.x+ .

The quick fix for this PR is to fix makefile to v1.54.1 : https://github.com/open-feature/open-feature-operator/blob/main/Makefile#L88-L89

config/crd/kustomization.yaml
```diff
@@ -4,19 +4,25 @@
...
 - patches/webhook_in_flagsourceconfigurations.yaml
+#- patches/webhook_in_featureflags.yaml
+#- patches/webhook_in_featureflagsources.yaml
 #+kubebuilder:scaffold:crdkustomizewebhookpatch
 
 # [CERTMANAGER] To enable cert-manager, uncomment all the sections with [CERTMANAGER] prefix.
 # patches here are for enabling the CA injection for each CRD
 - patches/cainjection_in_featureflagconfigurations.yaml
 - patches/cainjection_in_flagsourceconfigurations.yaml
+#- patches/cainjection_in_featureflags.yaml
```
thisthat, line 24: why is the CA injection disabled for this version?

odubajDT: was the default option when creating CRDs and not creating dedicated controllers for them. If we need it in the future, we can enable it. For now I would say it's ok

odubajDT:
> > > Approved, but I noticed [this](https://github.com/open-feature/open-feature-operator/pull/535/commits/b188fe6dc19d261ff21f238875f46d84a534283d#r1381656765) (just a question).
> > > Also I'm noticing a lint error in the CI which almost make it seems like a dependency might require a higher go version or something: https://github.com/open-feature/open-feature-operator/actions/runs/6730203336/job/18292441905?pr=535#step:6:493
> > 
> > 
> > Thank you!
> > The error is present also in the main branch and is not related to this PR. I would say it should be resolved in a separate PR. I can have a look at it.
> 
> The error comes with the usage of golangci-lint@latest - From version [v1.54.2](https://github.com/golangci/golangci-lint/releases/tag/v1.54.2) onwards, there's a transitive dependency which require Go version 1.20.x+ .
> 
> The quick fix for this PR is to fix makefile to v1.54.1 : https://github.com/open-feature/open-feature-operator/blob/main/Makefile#L88-L89

I already started working on a proper fix by moving the check to a separate workflow and also using a dedicated golangci-lint action for it https://github.com/open-feature/open-feature-operator/pull/538 

Hopefully it will be ready soon

PR #541, opened 2023-11-14
feat: release APIs and Operator independently

thisthat:
You can see the tags created in my fork here: https://github.com/thisthat/open-feature-operator/tags

For the Operator, the release works as previously creating containers and publishing them to a registry: https://github.com/thisthat/open-feature-operator/actions/runs/6861654249

For the APIs, only a tag is created: https://github.com/thisthat/open-feature-operator/actions/runs/6861600403

This PR covers only the release process. We should also move all go files into two different go modules.
This entails the following:
1. Creating a `go.mod` file under the `apis` folder
2. Remove dependency from `utils` in the new `apis` pkg (it is used in `alpha1` and `alpha2`)


.release-please-manifest.json
```diff
@@ -1,3 +1,4 @@
 {
-  ".": "0.2.36"
+  ".": "0.2.36",
+  "apis": "0.2.36"
```
thisthat, line 3: we should decide what should be the first version of the APIs package.

toddbaert: I think same version as the current package makes sense...?

release-please-config.json
```diff
@@ -1,11 +1,28 @@
...
     ".": {
       "release-type": "go",
+      "package-name": "operator",
+      "draft": false,
       "prerelease": false,
       "bump-minor-pre-major": true,
       "bump-patch-for-minor-pre-major": true,
+      "exclude-paths": [
+        "apis"
+      ],
```
toddbaert, line 25: :+1:

toddbaert:
I'm 100% on board with this.

release-please-config.json
beeme1mr: I didn't know this was an option.

.release-please-manifest.json
beeme1mr: This will slightly change how the release process works. @toddbaert and I have dealt with this enough times that it shouldn't be an issue here.

thisthat:
Nice, then I'll work on this PR making it fully working ;)

thisthat:
To continue with this PR, I would need a tag for the `/apis` package. @toddbaert maybe we can have a sync and add that manually for the first release of the APIs :)

PR #550, opened 2023-11-17
feat: restricting sidecar image and tag setup

### LINKED ISSUE #517: Consider restricting FlagSourceConfiguration CRD

odubajDT:
Fixes #517

## Changes

- remove the possibility to set sidecar `image` and `tag` via FFS CR, now it's only possible to set it via operator env variables
- unify env variable processing into a single place (`main.go`) and use standardized struct (`envconfig`) to read the env variables and potentially set the default values
- use latest api
- regenerate CRD manifests and CRD documentation
- adapt tests


Kavindu-Dodan (approved):

main.go
```diff
@@ -20,8 +20,10 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"log"
 	"os"
 
+	"github.com/kelseyhightower/envconfig"
```
Kavindu-Dodan, line 26: +1 for this

toddbaert (approved):

toddbaert:
Thanks!

PR #573, opened 2023-12-04
chore: add helm migration section

Kavindu-Dodan:
## This PR

Adds helm migration section and add refernce to `FlagSourceConfiguration` renaming

### DIFF
--- a/docs/v1beta_migration.md
+++ b/docs/v1beta_migration.md
@@ -76,3 +76,5 @@ We recommend following migration steps,
 3. Install upgraded custom resources
 4. Update annotation of your workloads to the latest supported version
 
+If you have used `flagd-proxy` provider, then you have to upgrade the image used by the `flagd-proxy` deployment. 
+For this, please edit the deployment of `flagd-proxy` to version [v0.3.1](https://github.com/open-feature/flagd/pkgs/container/flagd-proxy/152333134?tag=v0.3.1) or above. 
--- a/docs/v1beta_migration.md
+++ b/docs/v1beta_migration.md
@@ -13,8 +13,9 @@ Consider below example with diff from old version to new version,
 
 ```diff
 - apiVersion: core.openfeature.dev/v1alpha3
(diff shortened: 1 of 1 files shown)


docs/v1beta_migration.md
```diff
@@ -60,3 +64,15 @@ spec:
...
+## Migration on helm
+
+The new operator no longer support older API versions. Because of this, you need to plan your upgrade carefully.
+
+We recommend following migration steps,
+
+1. Remove all the old custom resources while running the older version of the operator
+2. Update the operator to the latest version
+3. Install upgraded custom resources
+4. Update annotation of your workloads to the latest supported version
```
toddbaert, line 77: I haven't actually tested this. What's the impact of removing the old CRs? while OFO is running? Does that mean deleting the instances as well? Can you delete a CRD while there are still CRs?

If there's any "gotchas" here, are they worth noting? Or do things work pretty smoothly?

Kavindu-Dodan: > What's the impact of removing the old CRs

For K8s syncs, resource removal doesn't do anything per current logic. And from operator logic, per my knowledge there's no deployment modification for CR removal. So we do not remove any flagd instance (this is what I observed as well)

> Can you delete a CRD while there are still CRs

You can but this lead to inconsistency. You can't remove CRs  without force as K8s no longer understand the API of the CR (API removed with CRD)

> If there's any "gotchas" here, are they worth noting

API removal and CRD changes are tricky. And given we do not have backward compatibility, best we can propose is to clean up CRs prior to OFO upgrade. And then update workloads with new annotations, which trigger OFO for new configurations in the system.

Kavindu-Dodan: I found one gotcha, the upgrade does not bump the flagd-proxy version. This has to be manually updated after the new OFO deployment by updating the deployment. I will document this

beeme1mr (approved):

Issue #579, opened 2023-12-06
OpenFeature is disabled?


agardnerIT:
In the same investigation as open-feature/open-feature-operator#575 and open-feature/flagd#1063 

```
2023-12-06T23:11:23Z	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/mutate-v1-pod", "UID": "8f021e26-6d9b-47f2-90f2-588236ec388d", "kind": "/v1, Kind=Pod", "resource": {"group":"","version":"v1","resource":"pods"}}2023-12-06T23:11:23Z	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/mutate-v1-pod", "code": 200, "reason": "OpenFeature is disabled", "UID": "8f021e26-6d9b-47f2-90f2-588236ec388d", "allowed": true}
```

Also noticed:

```
2023-12-06T23:10:31Z	ERROR	setup	podMutator backfill permissions error	{"error": "Index with name field:metadata.annotations.openfeature.dev/allowkubernetessync does not exist"}
```

OFO v0.5.2 deployed w/ Argo (Helm) + flagd proxy.

## Application and Flag Config

```yaml
apiVersion: core.openfeature.dev/v1beta1
kind: FeatureFlag
metadata:
  name: sample-flags
  namespace: openfeatureplayground-team70-preprod-cd
  labels:
    app: open-feature-demo
spec:
  flagSpec:
    flags:
      new-welcome-message:
        state: ENABLED
        variants:
          'on': true
          'off': false
        defaultVariant: 'off'
---
apiVersion: core.openfeature.dev/v1beta1
kind: FeatureFlagSource
metadata:
  name: flag-source
  namespace: openfeatureplayground-team70-preprod-cd
spec:
  sources:
  - source: openfeatureplayground-team70-preprod-cd/sample-flags
    provider: flagd-proxy
---
apiVersion: argoproj.io/v1alpha1
kind: Rollout
metadata:
  name: "openfeatureplayground-team70"
  namespace: "openfeatureplayground-team70-preprod-cd"
  labels:
    dt.owner: "team70"
spec:
  replicas: 2
  strategy:
    canary:
      steps:
      - setWeight: 50
      - pause: {duration: 5s}
      - setWeight: 100
  revisionHistoryLimit: 0
  selector:
    matchLabels:
      app.kubernetes.io/name: openfeaturedemo
  template:
    metadata:
      annotations:
[...]

PR #582, opened 2023-12-07
fix: create index for pod annotation path for allowkubernetessync annotation instead of deployment

odubajDT:
Fixes: #579

### DIFF
--- a/main.go
+++ b/main.go
@@ -137,14 +137,14 @@ func main() {
 	if err := mgr.GetFieldIndexer().IndexField(
 		context.Background(),
 		&corev1.Pod{},
-		fmt.Sprintf("%s/%s", common.OpenFeatureAnnotationPath, common.AllowKubernetesSyncAnnotation),
+		fmt.Sprintf("%s/%s", common.PodOpenFeatureAnnotationPath, common.AllowKubernetesSyncAnnotation),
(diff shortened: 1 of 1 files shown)


Kavindu-Dodan (approved):

toddbaert:
Can we add some kind of test for this? e2e or otherwise?

odubajDT:
> Can we add some kind of test for this? e2e or otherwise?
> 
> Another question I have is that `wrote response {"webhook": "/mutate-v1-pod", "code": 200, "reason": "OpenFeature is disabled", "UID": "8f021e26-6d9b-47f2-90f2-588236ec388d", "allowed": true}` (meaning the `openfeature/enabled` annotation isn't found) is also reported in the issue. Is that related or is that a totally different issue (I think it's different)? Note my comment [here](https://github.com/open-feature/open-feature-operator/issues/579#issuecomment-1845375414).

Hey, to be honest I don't know. On one side, if the OF is disabled, the mutating webhook should never kick in and therefore we should not have any problems with backfilling the permissions. Here we have both, maybe Adam can give us more context here? One theory might be that during the rollout, there was a certain time, where the namespace was created, but not annotated for OF yet. There we might see this warning (OF disabled). After it's annotated, we saw the errors from mutating webhook. Just thinking out loud... WDYT?

PR #587, opened 2023-12-15
chore: Remove metrics-port flag/usage from flagdproxy startup

### LINKED ISSUE #586: Remove deprecated (and soon to be removed) --metrics-port usage from flagd-proxy startup.
Remove [these instances](https://github.com/search?q=repo%3Aopen-feature%2Fopen-feature-operator+metrics-port&type=code) of deprecated `--metrics-port` and use `--management-port` instead.

See related flagd change.

:warning: this MUST be done before consuming the next release of flagd-proxy

webstradev:
## This PR

- Remove the usage of the metrics-port flag and and the corresponding tests.
This flag was deprecated and is now removed from flagd so it needs to be removed here before using the new version of flagd.

### Related Issues

Closes #586

### DIFF
--- a/common/flagdproxy/flagdproxy.go
+++ b/common/flagdproxy/flagdproxy.go
@@ -120,7 +120,7 @@ func (f *FlagdProxyHandler) newFlagdProxyManifest(ownerReferences []metav1.Owner
 	replicas := int32(1)
 	args := []string{
 		"start",
-		"--metrics-port",
+		"--management-port",
 		fmt.Sprintf("%d", f.config.ManagementPort),
 	}
 	if f.config.DebugLogging {
@@ -165,7 +165,7 @@ func (f *FlagdProxyHandler) newFlagdProxyManifest(ownerReferences []metav1.Owner
 									ContainerPort: int32(f.config.Port),
 								},
 								{
-									Name:          "metrics-port",
+									Name:          "management-port",
 									ContainerPort: int32(f.config.ManagementPort),
 								},
 							},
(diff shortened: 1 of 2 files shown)


common/flagdproxy/flagdproxy.go
```diff
@@ -120,7 +120,6 @@ func (f *FlagdProxyHandler) newFlagdProxyManifest(ownerReferences []metav1.Owner
 	replicas := int32(1)
 	args := []string{
 		"start",
-		"--metrics-port",
```
toddbaert, line 123: Hey @webstradev ! I think you'll want to change this to `--management-port`, instead of removing it entirely.

We're basically using the port for more than just metrics now, so the name was misleading - but we still want the same configurability.

toddbaert: In fact, I think we have a similar regression [here](https://github.com/open-feature/open-feature-operator/blob/a6fa04f590ad4ad6779ce85f4fc167b59f1b17a7/common/flagdinjector/flagdinjector.go#L394) where we also seem to be configuring the ports for the container, but failing to pass the port to the flagd startup.

webstradev: Gotcha, I'll sort that tonight

webstradev: @toddbaert I have updated the two usages ofmetrics port. 

I'm uncertain about the third place you mentioned, because if I add the --management-port flag to that start command you highlighted there then a whole bunch of tests start failing.
Happy to look into that one further if you think it is necesary though.

toddbaert: OK I will investigate that and make a new issue.

toddbaert (approved):

beeme1mr (approved):

Issue #583, opened 2023-12-09
Auto-upgrade flagd-proxy version with OFO upgrades


toddbaert:
Currently, when a new version of OFO is installed, we don't update [flagd-proxy](https://github.com/open-feature/flagd/tree/main/flagd-proxy) . In deployments where it's active, OFO should gracefully update this component. Note this doesn't necessarily have to be part of the actual upgrade process - it might be something OFO could do at runtime.

Tasks:  
 
- [x] implement graceful upgrade (or downgrade)
  - [x] only upgrade/downgrade if the specified flagd-proxy version does not match running version (if not specified, that means the default version associated with the OFO release should be used) 
  - [x] no downtime
  - [x] clear messaging for admins if something goes wrong
- [x] tests
- [ ] document this feature (especially the version behavior above) and possible "gotchas", if any

PR #596, opened 2024-01-15
feat: auto-upgrade flagd-proxy with OFO upgrades

odubajDT:
Fixes: #583


common/flagdproxy/flagdproxy.go
```diff
@@ -62,32 +64,44 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
 	return f.config
 }
 
+func (f *FlagdProxyHandler) CreateObject(ctx context.Context, obj client.Object) error {
+	return f.Client.Create(ctx, obj)
+}
+
+func (f *FlagdProxyHandler) UpdateObject(ctx context.Context, obj client.Object) error {
+	return f.Client.Update(ctx, obj)
+}
```
Kavindu-Dodan, line 73: Both can be non-exported methods. And you may consider inlining these operations

common/flagdproxy/flagdproxy.go
```diff
@@ -206,3 +223,14 @@ func (f *FlagdProxyHandler) getOwnerReference(ctx context.Context) (metav1.Owner
 	}, nil
 
 }
+
+func (f *FlagdProxyHandler) getOwnerReferences(ctx context.Context) []metav1.OwnerReference {
```
Kavindu-Dodan, line 227: can't we combine this functionality with `getOwnerReference(ctx context.Context) (metav1.OwnerReference, error)` ?

Kavindu-Dodan (approved):
Left minor improvement suggestions, otherwise looks good

chart/open-feature-operator/README.md
```diff
@@ -45,6 +45,14 @@ helm repo update
 helm upgrade --install open-feature-operator openfeature/open-feature-operator
 ```
 
+**Note:**
+If you have used `flagd-proxy` provider and upgrading to OFO version `v0.5.4` or higher,
+`flagd-proxy` will be automatically upgraded to the lastest supported version by the `open-feature-operator`.
```
beeme1mr, line 50:
```suggestion
`flagd-proxy` will be automatically upgraded to the latest supported version by the `open-feature-operator`.
```

chart/open-feature-operator/README.md
```diff
@@ -45,6 +45,14 @@ helm repo update
 helm upgrade --install open-feature-operator openfeature/open-feature-operator
 ```
 
+**Note:**
```
beeme1mr, line 48: Could you please use the GitHub note syntax?

https://github.com/orgs/community/discussions/16925

docs/installation.md
```diff
@@ -28,6 +28,14 @@ helm upgrade --install openfeature openfeature/open-feature-operator
 helm upgrade --install openfeature openfeature/open-feature-operator
 ```
 
+**Note:**
```
beeme1mr, line 31: Same comment as above

docs/installation.md
```diff
@@ -28,6 +28,14 @@ helm upgrade --install openfeature openfeature/open-feature-operator
...
+If you have used `flagd-proxy` provider and upgrading to OFO version `v0.5.4` or higher,
+`flagd-proxy` will be automatically upgraded to the lastest supported version by the `open-feature-operator`.
```
beeme1mr, line 33:
```suggestion
`flagd-proxy` will be automatically upgraded to the latest supported version by the `open-feature-operator`.
```

docs/v1beta_migration.md
```diff
@@ -76,5 +76,10 @@ We recommend following migration steps,
 3. Install upgraded custom resources
 4. Update annotation of your workloads to the latest supported version
 
-If you have used `flagd-proxy` provider, then you have to upgrade the image used by the `flagd-proxy` deployment. 
-For this, please edit the deployment of `flagd-proxy` to version [v0.3.1](https://github.com/open-feature/flagd/pkgs/container/flagd-proxy/152333134?tag=v0.3.1) or above. 
+If you have used `flagd-proxy` provider, then you have to upgrade the image used by the `flagd-proxy` deployment.
+For this, please edit the deployment of `flagd-proxy` to version [v0.3.1](https://github.com/open-feature/flagd/pkgs/container/flagd-proxy/152333134?tag=v0.3.1) or above.
+
+**Note:**
```
beeme1mr, line 82: Same comment as above.

odubajDT:
> Left minor improvement suggestions, otherwise looks good

Adapted, thank you!

chart/open-feature-operator/README.md
```diff
@@ -45,6 +45,13 @@ helm repo update
 helm upgrade --install open-feature-operator openfeature/open-feature-operator
 ```
 
+> [!NOTE]
+> If you have used `flagd-proxy` provider and upgrading to OFO version `v0.5.4` or higher,
+`flagd-proxy` will be automatically upgraded to the latest supported version by the `open-feature-operator`.
```
thisthat, line 50:
```suggestion
> If you upgrade to OFO `v0.5.4` or higher while using a `flagd-proxy` provider, the instance of
`flagd-proxy` will be automatically upgraded to the latest supported version by the `open-feature-operator`.
```

chart/open-feature-operator/README.md
```diff
@@ -45,6 +45,13 @@ helm repo update
...
+This upgrade will also consider your current `FeatureFlagSource` configuration and adapt
+the `flagd-proxy` Deployment accordingly.
```
thisthat, line 52: for "this" you are talking about `flagd-proxy` or OFO?

common/flagdproxy/flagdproxy.go
```diff
@@ -28,6 +29,7 @@ type FlagdProxyHandler struct {
 	Log    logr.Logger
 }
 
+type CreateUpdateFunc func(ctx context.Context, obj client.Object) error
```
thisthat, line 32: nit:
```suggestion
type CreateUpdateFunc func(ctx context.Context, obj client.Object) error

```

common/flagdproxy/flagdproxy.go
```diff
@@ -62,32 +64,44 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
...
-	if err != nil {
-		f.Log.Error(err, "unable to create owner reference for open-feature-operator, not appending")
-	} else {
-		ownerReferences = append(ownerReferences, ownerReference)
-	}
-
+func (f *FlagdProxyHandler) deployFlagdProxy(ctx context.Context, ff CreateUpdateFunc, deployment *appsV1.Deployment, service *corev1.Service) error {
 	f.Log.Info("deploying the flagd-proxy")
-	if err := f.Client.Create(ctx, f.newFlagdProxyManifest(ownerReferences)); err != nil && !errors.IsAlreadyExists(err) {
+	if err := ff(ctx, deployment); err != nil && !errors.IsAlreadyExists(err) {
```
thisthat, line 100: I don't like `ff` but I don't have a better name 😢 
Maybe `editf`?

common/flagdproxy/flagdproxy.go
```diff
@@ -178,31 +192,46 @@ func (f *FlagdProxyHandler) newFlagdProxyManifest(ownerReferences []metav1.Owner
...
+		return false, nil, err
 	}
-	// exists, at least one replica ready, no error
-	return true, nil
+	return true, d, nil
 }
 
-func (f *FlagdProxyHandler) getOwnerReference(ctx context.Context) (metav1.OwnerReference, error) {
+func (f *FlagdProxyHandler) isFlagdProxyUpToDate(old, new *appsV1.Deployment) bool {
+	return reflect.DeepEqual(old.Spec, new.Spec)
```
thisthat, line 210: it would be interesting in the future to allow fix releases (basically, pin only major.minor) of flagd w/o the need of a new OFO version

common/flagdproxy/flagdproxy.go
```diff
@@ -62,32 +64,44 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
...
 
+func (f *FlagdProxyHandler) createObject(ctx context.Context, obj client.Object) error {
+	return f.Client.Create(ctx, obj)
+}
+
+func (f *FlagdProxyHandler) updateObject(ctx context.Context, obj client.Object) error {
+	return f.Client.Update(ctx, obj)
+}
+
 func (f *FlagdProxyHandler) HandleFlagdProxy(ctx context.Context) error {
```
thisthat, line 75: what should we do if a user deploys manually a `flagd-proxy` instance using the same name and in the same namespace but the `.spec` differs, i.e., the version is different? Currently, we just update it but IMHO we should look at the `managed-by` label. I am thinking about the case where there is a GitOps system in place (eg Argo) that deploys the proxy to a different version comparing to what OFO expects. This would trigger a continuous loop between OFO changing the deployment and Argo changing it back 🤔

common/flagdproxy/flagdproxy.go
Kavindu-Dodan: yeah, I thought about this too. To compare versions, we need to go deep into the deployment content (to the container level). Besides, I think we are safe with renovate and releases to make sure this fails only when there're upgrades.

common/flagdproxy/flagdproxy.go
odubajDT: yes that's true; not sure if `managed-by` label is set when installing from manifests; but we can definitely look at the `ownerReferences`. This should be set all the time -> Here we need to adapt the behavior a little bit (if we cannot fetch OFO Deployment -> do not create/update `flagd-proxy`)

WDYT?

common/flagdproxy/flagdproxy.go
odubajDT: But there are also other use-cases when we should update, let's imagine a user changes `FeatureFlagSource`, or env variable value for OFO Deployment (let's say `management-port`) -> this needs to change the deployment of flagd-proxy and adapt it according to the configuration. Therefore I think it's not enough to check for the image, as the flagd-proxy deployment needs to be adapted even if other configuration changes. This was the reason I went for the comparison of the whole spec.

thisthat: I think there is a misunderstanding here @odubajDT :) 
What I was thinking for the future, after flagd is at v1.x.y, is to allow old and new `.spec` fields] to differ only in the `image` field for the `tag` part. Since flagd follows semantic versioning, we would have a version schema `x.y.z` and we can check that new`z` is >= than old `z`. This way, users can patch their flagd-proxy manually to have a quick fix if smth is off without waiting for the related OFO release :) 
I hope this clarifies my comment :)

common/flagdproxy/flagdproxy.go
thisthat: `ownerReferences` is meant for GC, I don't know if it would fit this use case. Furthermore, `ownerReferences` would differ between two OFO versions, name and uid fields are not preserved between updates.

odubajDT: adapted according to description

toddbaert: I think this is an interesting point... either a label, or maybe a special annotation?

odubajDT: adapted it to use a label to check if the flagd-proxy deployment was deployed by OFO, can we maybe align here on the name of the label? @toddbaert @thisthat @Kavindu-Dodan ?

I would suggest 

`openfeature.dev/deployed: "true"`

WDYT?

odubajDT: Agreed with @thisthat to used `managed-by` k8s annotation, adapted according to this

Kavindu-Dodan: I think the new update to use `managed-by` is a great improvement 👍

thisthat (approved):
LGTM, I only left a minor open suggestion for logging :) thanks @odubajDT for the contribution 🙌

common/flagdproxy/flagdproxy.go
odubajDT: @toddbaert if you agree with the solution, we should be ready to merge

Kavindu-Dodan (approved):

beeme1mr (approved):

PR #607, opened 2024-02-20
chore: regex to match all go files

Kavindu-Dodan:
## This PR

Adds go files for regex pattern


toddbaert (approved):

toddbaert:
Ya I think this is a good idea

PR #622, opened 2024-04-23
feat: introduce validating webhook for FeatureFlag CR

### LINKED ISSUE #597: Add targeting validation
With the addition of a [rigorous targeting schema](https://flagd.dev/schema/v0/targeting.json) for flagd, we can validate targeting rules. We should validate CRDs with these.

Tasks:

- [x] add JSON schema validation for RULES only (we don't need to validate entire flag configs, just rules)
  - [x] clear error messaging to admins when this validation fails
- [x] add option arg to disable validation in OFO (should be on by default)
- [x] associated tests

odubajDT:
## This PR

This PR needs to have a follow-up PR, which updates the `apis` module in the `go.mod` file to actually make the changes in the `/apis` folder visible to the rest of the operator logic. This PR can be merged.

In the current state, some tests might be failing, green pipeline is visible for the previous commits (not counting the latest one).

### Related Issues

Fixes #597


apis/core/v1beta1/featureflag_types.go
```diff
@@ -31,14 +31,19 @@ type FeatureFlagSpec struct {
 }
 
 type FlagSpec struct {
-	Flags map[string]Flag `json:"flags"`
+	Flags `json:",inline"`
```
odubajDT, line 34: This change was needed due to marshaling of the structure to json. If we want to have the json in the format:

```
{
    "flags": {...}
}
```

The parameter needs to have a separate struct. If we would leave it in the original way, we will be missing the root "flags" in the marshalled json.

This is not a breaking change. It's just a change of the internal representation of the CR. This change is not visible to the user.

apis/core/v1beta1/featureflag_webhook_test.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+func Test_validateFeatureFlagTargeting(t *testing.T) {
+	tests := []struct {
+		name    string
+		in      Flags
+		wantErr bool
+	}{
+		{
+			name: "happy path",
```
odubajDT, line 17: Potentially more unit tests can be added from [here](https://github.com/open-feature/flagd-schemas/blob/main/json/test/negative/invalid-flagd-props.json), but in my eyes, more tests are testing the schema, not the validation webhook itself

apis/core/v1beta1/schema/flags.json
```diff
@@ -0,0 +1,185 @@
+{
```
odubajDT, line 1: Open question:

Should the schemas be added here as:

- url link
- local copy (as it is now)
- as a sub-repository

Kavindu-Dodan: +1 to have them as git submodels. This is how we do it many other repositories.

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,106 @@
...
+func validateFeatureFlagFlags(flags Flags) error {
+	b, err := json.Marshal(flags)
+	if err != nil {
+		return err
+	}
+
+	documentLoader := gojsonschema.NewStringLoader(string(b))
+	schemaLoader := gojsonschema.NewSchemaLoader()
+	schemaLoader.AddSchemas(gojsonschema.NewStringLoader(TargetingSchema))
+	compiledSchema, err := schemaLoader.Compile(gojsonschema.NewStringLoader(FlagsScheme))
```
Kavindu-Dodan, line 89: minor - I know we don't do this in flagd, but I think compiled schema is reusable. If so we should reuse it.

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,106 @@
...
+//go:embed schema/targeting.json
+var TargetingSchema string
+
+//go:embed schema/flags.json
+var FlagsScheme string
+
+// log is for logging in this package.
+var featureflaglog = logf.Log.WithName("featureflag-resource")
+
+func (r *FeatureFlag) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
Kavindu-Dodan, line 41: nit - `ff` for receiver (as suggested by my ide)

main.go
```diff
@@ -181,6 +181,13 @@ func main() {
 		os.Exit(1)
 	}
 
+	if env.FlagsValidationEnabled {
+		if err = (&corev1beta1.FeatureFlag{}).SetupWebhookWithManager(mgr); err != nil {
+			setupLog.Error(err, "unable to create webhook", "webhook", "FeatureFlag")
```
Kavindu-Dodan, line 186: nit 
```suggestion
			setupLog.Error(err, "unable to create the validation webhook for FeatureFlag CRD", "webhook", "FeatureFlag")
```

Kavindu-Dodan:
@odubajDT looks good 🙌

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,106 @@
...
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-core-openfeature-dev-v1beta1-featureflag,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.openfeature.dev,resources=featureflags,verbs=create;update,versions=v1beta1,name=vfeatureflag.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &FeatureFlag{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *FeatureFlag) ValidateCreate() error {
+	featureflaglog.Info("validate create", "name", r.Name)
```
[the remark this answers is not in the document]
odubajDT: The error messages are written out on stdout when applying malformed CR, so it would be doubling the output, these logs I find quite informal when debugging if the validation webhook kicked in, they were quite useful for s in the past in other projects

apis/core/v1beta1/featureflag_webhook.go
odubajDT: I was thinking about this as well, but the only solution might be to to store it to a global var, which I would like to avoid. Webhook is a one-time execution on a single CR create/update. Not sure how it works behind the scenes, but we may get some weird errors when starting to store things in global vars.

Or did you had another solution in mind?

apis/core/v1beta1/schema/flags.json
odubajDT: Experiencing problems to initialize the submodule and adding it to the built image, do you have an idea how this should be done properly? I am unable to make it work.

apis/core/v1beta1/featureflag_webhook.go
Kavindu-Dodan: The type/schema registration happens here [1], and here we simply create a struct. But yeah, I was also unable to find any details of the impact of having non API bound fields 

[1] - https://github.com/open-feature/open-feature-operator/blob/main/apis/core/v1beta1/featureflag_types.go#L86

apis/core/v1beta1/schema/flags.json
odubajDT: ok, adapted!

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+	schema "github.com/open-feature/flagd-schemas/json"
+	"github.com/xeipuuv/gojsonschema"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var featureflaglog = logf.Log.WithName("featureflag-resource")
```
bacherfl, line 34: nit:

```suggestion
var featureFlagLog = logf.Log.WithName("featureflag-resource")
```

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+		for _, desc := range result.Errors() {
+			err = fmt.Errorf(err.Error() + desc.Description() + "\n")
+		}
+	}
+	return err
+}
+
+func initSchemas() (*gojsonschema.Schema, error) {
+	schemaLoader := gojsonschema.NewSchemaLoader()
+	schemaLoader.AddSchemas(gojsonschema.NewStringLoader(schema.TargetingSchema))
```
bacherfl, line 104:
```suggestion
		if err := schemaLoader.AddSchemas(gojsonschema.NewStringLoader(schema.TargetingSchema)); err != nil {
		return nil, err
	}
```

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,111 @@
...
+	if !result.Valid() {
+		err = fmt.Errorf("")
+		for _, desc := range result.Errors() {
+			err = fmt.Errorf(err.Error() + desc.Description() + "\n")
+		}
+	}
+	return err
+}
+
+func initSchemas() (*gojsonschema.Schema, error) {
```
bacherfl, line 102: can we do this with a `sync.Once` so we don't have to compile the schema at every validation?

apis/core/v1beta1/featureflag_webhook_test.go
```diff
@@ -0,0 +1,167 @@
...
+							"none": "none"}
+						`),
+						DefaultVariant: "none",
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "flactional invalid bucketing",
```
bacherfl, line 63:
```suggestion
			name: "fractional invalid bucketing",
```

go.mod
```diff
@@ -75,4 +79,7 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace golang.org/x/net => golang.org/x/net v0.24.0
+replace (
+	github.com/open-feature/open-feature-operator/apis => github.com/odubajDT/open-feature-operator/apis v0.0.0-20240426164950-7aa22f92ccec
```
bacherfl, line 83: i guess this can be removed

odubajDT: this PR should not be merged, it is only for reviews (to this replace is OK)

bacherfl (approved):

go.mod
```diff
@@ -75,4 +79,7 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace golang.org/x/net => golang.org/x/net v0.24.0
+replace (
+	github.com/open-feature/open-feature-operator/apis => github.com/odubajDT/open-feature-operator/apis v0.0.0-20240502135831-c4a6ccf893a2
+	golang.org/x/net => golang.org/x/net v0.24.0
+)
```
odubajDT, line 85:
```suggestion
replace golang.org/x/net => golang.org/x/net v0.24.0
```

go.sum
```diff
@@ -219,12 +219,14 @@ github.com/mwitkow/go-conntrack v0.0.0-20161129095857-cc309e4a2223/go.mod h1:qRW
 github.com/mwitkow/go-conntrack v0.0.0-20190716064945-2f068394615f/go.mod h1:qRWi+5nqEBWmkhHvq77mSJWrCKwh8bxhgT7d/eI7P4U=
 github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e h1:fD57ERR4JtEqsWbfPhv4DMiApHyliiK5xCTNVSPiaAs=
 github.com/niemeyer/pretty v0.0.0-20200227124842-a10e7caefd8e/go.mod h1:zD1mROLANZcx1PVRCS0qkT7pwLkGfwJo4zjcN/Tysno=
+github.com/odubajDT/open-feature-operator/apis v0.0.0-20240502135831-c4a6ccf893a2 h1:j1vjWqNuCkAHs7UPCYgfAE1nZH8VnsR2p3w8WAkcSYw=
+github.com/odubajDT/open-feature-operator/apis v0.0.0-20240502135831-c4a6ccf893a2/go.mod h1:BLFkIDsj+coW30XxILSY2X++vVniRTyWozzONuVKh5U=
```
odubajDT, line 223:
```suggestion
```

apis/core/v1beta1/featureflag_webhook.go
odubajDT: Already a singleton initialization is happening here. Hope this fits your ideas

apis/core/v1beta1/featureflag_webhook.go
```diff
@@ -0,0 +1,117 @@
...
+		Complete()
+}
+
+//+kubebuilder:webhook:path=/validate-core-openfeature-dev-v1beta1-featureflag,mutating=false,failurePolicy=fail,sideEffects=None,groups=core.openfeature.dev,resources=featureflags,verbs=create;update,versions=v1beta1,name=vfeatureflag.kb.io,admissionReviewVersions=v1
+
+var _ webhook.Validator = &FeatureFlag{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (ff *FeatureFlag) ValidateCreate() error {
+	featureFlagLog.Info("validate create", "name", ff.Name)
```
toddbaert, line 51: Is it standard to have these as `Info` level? If it is, that's fine, I'm just curious. I seems like more of a `Debug` thing... but maybe this is standard for better logging of resource changes? I suspect you know better than me.

apis/core/v1beta1/featureflag_webhook_test.go
toddbaert: Agreed. I think as long as you test 1 failure and 1 success that's sufficient for OFO. This is more than enough.

toddbaert (approved):
LGTM. Next week we'll force merge and do the update you mention.

apis/core/v1beta1/featureflag_webhook.go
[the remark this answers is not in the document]
toddbaert: I think having log traceability here probably makes sense... it's not like it will hurt performance or anything at the rate we'd expect these to change.

 I will defer to people who know more about debugging operators though.

apis/core/v1beta1/featureflag_webhook.go
odubajDT: This part was auto generated by kubebuilder. I don't have a strong opinion for this, but it can be changed to debug mode if you like

toddbaert: Use your judgement, I just wanted to point it out.

PR #632, opened 2024-05-07
chore: PoC new CRD for in-process evaluation [WIP]

odubajDT:


apis/core/v1beta1/common/common.go
```diff
@@ -12,6 +12,36 @@ const (
 	SyncProviderFlagdProxy SyncProviderType = "flagd-proxy"
 )
 
+const (
+	MetricPortEnvVar            string = "MANAGEMENT_PORT"
```
bacherfl, line 16: to keep it consistent I think we could also rename this to `ManagementPortEnvVar`:

```suggestion
	ManagementPortEnvVar            string = "MANAGEMENT_PORT"
```

webhooks/pod_webhook.go
```diff
@@ -139,6 +159,30 @@ func (m *PodMutator) createFSConfigSpec(ctx context.Context, req admission.Reque
...
+	}
+
+	featureFlagSourceSpec := api.FeatureFlagInProcessConfigurationSpec{}
+
+	for _, fscName := range fscNames {
+		ns, name := utils.ParseAnnotation(fscName, req.Namespace)
+
+		fc, err := m.getFeatureFlagInProcessConfiguration(ctx, ns, name)
+		if err != nil {
+			m.Log.V(1).Info(fmt.Sprintf("FeatureFlagInProcessConfiguration could not be found for %s", fscName))
```
bacherfl, line 177: at this point the returned `err` can be either a `NotFound`, but also any other kind of error - should we distinguish between them here and also include the error in the log message? Also, let's include the namespace in the log to correctly relate the error to the request

webhooks/pod_webhook.go
```diff
@@ -74,10 +75,16 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
 		return admission.Denied("static or orphaned pods cannot be mutated")
 	}
 
-	if code, err := m.handleRPCEvaluation(ctx, req, annotations, pod); err != nil {
-		if code == 0 {
-			return admission.Denied(err.Error())
-		} else {
+	if shouldUseRPC(annotations) {
+		if code, err := m.handleRPCEvaluation(ctx, req, annotations, pod); err != nil {
+			if code == 0 {
```
bacherfl, line 80: to make if easier to understand the semantics of the `code` returned by this function, should we rather make it return `http.StatusForbidden` in this case? since this is also what will be set in `admission.Denied()`

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
```diff
@@ -0,0 +1,193 @@
...
+
+func init() {
+	SchemeBuilder.Register(&FeatureFlagInProcessConfiguration{}, &FeatureFlagInProcessConfigurationList{})
+}
+
+func (fc *FeatureFlagInProcessConfigurationSpec) Merge(new *FeatureFlagInProcessConfigurationSpec) {
+	if new == nil {
+		return
+	}
+	if len(new.EnvVars) != 0 {
```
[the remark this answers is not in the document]
odubajDT: you are right! I will adapt

beeme1mr (approved):

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
```diff
@@ -0,0 +1,195 @@
...
+
+	"github.com/open-feature/open-feature-operator/apis/core/v1beta1/common"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FeatureFlagInProcessConfigurationSpec defines the desired state of FeatureFlagInProcessConfiguration
+type FeatureFlagInProcessConfigurationSpec struct {
+	// Port defines the port to listen on, defaults to 8013
+	// +kubebuilder:default:=8013
```
thisthat, line 30: not related to this PR but should we move the default to 8080 on flagd since it's HTTP traffic? 🤔 @toddbaert @Kavindu-Dodan

apis/core/v1beta1/common/common.go
```diff
@@ -55,3 +89,28 @@ func FeatureFlagConfigurationId(namespace, name string) string {
...
+	for _, item := range input {
+		list[item.Name] = item
+	}
+	for _, val := range list {
+		out = append(out, val)
+	}
+	return out
+}
+
+func RemoveDuplicatesGeneric[T comparable](input []T) []T {
```
thisthat, line 105: should we just maintain the generic one? 🤔

webhooks/common.go
```diff
@@ -85,10 +85,23 @@ func NewFeatureFlagSourceSpec(env types.EnvConfig) *api.FeatureFlagSourceSpec {
 	}
 }
 
+func shouldUseRPC(annotations map[string]string) bool {
```
thisthat, line 88: nit since also in-process is over rpc
```suggestion
func shouldUseSidecar(annotations map[string]string) bool {
```

toddbaert:
I think we need to substantially rework OFO's documentation, including this. I think it's out of scope of this PR.

cc @Kavindu-Dodan @beeme1mr

webhooks/common.go
thisthat: I noticed that we have `handleRPCEvaluation` and I suggest changing to something more fitting like `handleSidecarInjection`

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
toddbaert: I might be confused but I still think this is gRPC traffic.

webhooks/common.go
toddbaert: I agree with this. I find that `handleRPCEvaluation` confusing, because at the time this runs, we're not evaluating anything.

webhooks/pod_webhook.go
```diff
@@ -90,6 +98,19 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
 
+func (m *PodMutator) handleInProcessEvaluation(ctx context.Context, req admission.Request, annotations map[string]string, pod *corev1.Pod) (int32, error) {
```
toddbaert, line 101: I'd recommend `handleInProcessConfiguration` and `handlePCConfiguration`

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
```diff
@@ -0,0 +1,195 @@
...
+
+	envs = append(envs, corev1.EnvVar{
+		Name:  common.EnvVarKey(fc.EnvVarPrefix, common.SocketPathEnvVar),
+		Value: fc.SocketPath,
+	})
+
+	envs = append(envs, corev1.EnvVar{
+		Name:  common.EnvVarKey(fc.EnvVarPrefix, common.OfflineFlagSourcePathEnvVar),
+		Value: fc.OfflineFlagSourcePath,
+	})
```
toddbaert, line 177: Should we support this? @Kavindu-Dodan @beeme1mr 

This is more of a non-prod thing, I think, right?

apis/core/v1beta1/common/common.go
```diff
@@ -12,6 +16,36 @@ const (
...
+	ManagementPortEnvVar        string = "MANAGEMENT_PORT"
+	PortEnvVar                  string = "PORT"
+	HostEnvVar                  string = "HOST"
+	TLSEnvVar                   string = "TLS"
+	SocketPathEnvVar            string = "SOCKET_PATH"
+	OfflineFlagSourcePathEnvVar string = "OFFLINE_FLAG_SOURCE_PATH"
+	SelectorEnvVar              string = "SOURCE_SELECTOR"
+	CacheEnvVar                 string = "CACHE"
+	CacheMaxSizeEnvVar          string = "MAX_CACHE_SIZE"
+	ResolverEnvVar              string = "RESOLVER"
```
toddbaert, line 29: This var (`FLAGD_RESOLVER`) needs to be set, especially in the in-process mode, but I don't see it used. cc @beeme1mr am I missing something?

toddbaert: On a related note, it looks like in Java we don't define this var: https://github.com/open-feature/java-sdk-contrib/blob/703830d5247b4a52a0b0767bf2d13a7e93582d00/providers/flagd/README.md?plain=1#L88

We need to do this to support auto-config of in-process flagd Java via OFO.

toddbaert: And looks like we have a bug (wrong var name) in dotnet: https://github.com/open-feature/dotnet-sdk-contrib/blob/38a59f01b4c740ddcfb69b68c8b79fb169e06ad4/src/OpenFeature.Contrib.Providers.Flagd/FlagdConfig.cs#L39

toddbaert: Fix for c#: https://github.com/open-feature/dotnet-sdk-contrib/pull/206

apis/core/v1beta1/common/common.go
odubajDT: this would be ideal, but if you consider `EnvVar` struct, we are not removing duplications (my bad, will adapt the naming). Let's take an example:

```yaml
- name: key1
   value: val1
- name: key1
   value: val2
```

These two items are not duplications, as the values are different, but in our use-case, we cannot have 2x the same key in `envVars`, therefore we need to store the key with the value only once. In this case taking the latest value, as this is the approach that is used when merging structs in OFO

apis/core/v1beta1/common/common.go
odubajDT: adapted

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
thisthat: I confused myself between ports 🤦 8015 is the one for OFREP and that's the one I was thinking to move to 8080

apis/core/v1beta1/featureflagsource_types.go
```diff
@@ -234,40 +214,45 @@ func (fc *FeatureFlagSourceSpec) ToEnvVars() []corev1.EnvVar {
...
-	if fc.LogFormat != defaultLogFormat {
+	if fc.LogFormat != common.DefaultLogFormat {
 		envs = append(envs, corev1.EnvVar{
-			Name:  common.EnvVarKey(fc.EnvVarPrefix, SidecarLogFormatEnvVar),
+			Name:  common.EnvVarKey(fc.EnvVarPrefix, common.LogFormatEnvVar),
 			Value: fc.LogFormat,
 		})
 	}
 
+	envs = append(envs, corev1.EnvVar{
```
toddbaert, line 252:
```suggestion
        // sets the FLAGD_RESOLVER var to "rpc" to configure the provider for RPC evaluation mode
	envs = append(envs, corev1.EnvVar{
```

Suggesting this comment because this is particularly important.

toddbaert (commented):
This looks good to me.

I want to run a manual test with a real workload to confirm everything works end-to-end, and then I'll approve.

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
Kavindu-Dodan: Sorry for late review, OFREP runs on 8016 . And gRPC sync service runs on 8015 [1]  (8013 is for remote evaluations).

Having dedicated ports is confusing but also gives users ability to control deployment at network level.

Anyway this port refers to gRPC sync port, so the default should be 8015 :) 

[1] - https://github.com/open-feature/flagd/blob/main/flagd/cmd/start.go#L41-L45

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
Kavindu-Dodan: yeah non-prod but a valid use-case so things can work offline

Kavindu-Dodan:
I will review this with a local run tomorrow 🤝

apis/core/v1beta1/common/common.go
thisthat: Then we should call it `RemoveDuplicatesSlice`, the generic aspect is forced by the compiler anyway via syntax :p

apis/core/v1beta1/common/common.go
```diff
@@ -55,3 +91,35 @@ func FeatureFlagConfigurationId(namespace, name string) string {
...
+func isEnvVarNamePresent(slice []corev1.EnvVar, item corev1.EnvVar) bool {
+	for _, i := range slice {
+		if i.Name == item.Name {
+			return true
+		}
+	}
+	return false
+}
+
+func RemoveDuplicatesGeneric[T comparable](input []T) []T {
```
odubajDT, line 114:
```suggestion
func RemoveDuplicatesFromSlice[T comparable](input []T) []T {
```

apis/core/v1beta1/common/common_test.go
```diff
@@ -34,3 +35,114 @@ func Test_FLagSourceConfiguration_FeatureFlagConfigurationId(t *testing.T) {
...
+	}
+	input2 := []int{
+		1, 2, 3, 4, 2,
+	}
+
+	require.Equal(t, RemoveDuplicatesGeneric(input1), []string{
+		"some", "input", "duplicate",
+	})
+
+	require.Equal(t, RemoveDuplicatesGeneric(input2), []int{
```
odubajDT, line 145:
```suggestion
	require.Equal(t, RemoveDuplicatesFromSlice(input1), []string{
		"some", "input", "duplicate",
	})

	require.Equal(t, RemoveDuplicatesFromSlice(input2), []int{
```

apis/core/v1beta1/common/common_test.go
```diff
@@ -34,3 +35,114 @@ func Test_FLagSourceConfiguration_FeatureFlagConfigurationId(t *testing.T) {
...
+			},
+		},
+	})
+	require.Contains(t, out3, corev1.EnvVar{
+		Name:  "key2",
+		Value: "val2",
+	})
+}
+
+func Test_RemoveDuplicateGenerics(t *testing.T) {
```
odubajDT, line 133:
```suggestion
func Test_RRemoveDuplicatesFromSlice(t *testing.T) {
```

apis/core/v1beta1/featureflagsource_types.go
```diff
@@ -197,9 +175,11 @@ func (fc *FeatureFlagSourceSpec) Merge(new *FeatureFlagSourceSpec) {
 	}
 	if len(new.EnvVars) != 0 {
 		fc.EnvVars = append(fc.EnvVars, new.EnvVars...)
+		fc.EnvVars = common.RemoveDuplicateEnvVars(fc.EnvVars)
 	}
 	if len(new.SyncProviderArgs) != 0 {
 		fc.SyncProviderArgs = append(fc.SyncProviderArgs, new.SyncProviderArgs...)
+		fc.SyncProviderArgs = common.RemoveDuplicatesGeneric[string](fc.SyncProviderArgs)
```
odubajDT, line 182:
```suggestion
		fc.SyncProviderArgs = common.RemoveDuplicatesFromSlice[string](fc.SyncProviderArgs)
```

webhooks/pod_webhook.go
```diff
@@ -74,10 +76,16 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
-		} else {
+	if shouldUseSidecar(annotations) {
+		if code, err := m.handleRPCConfiguration(ctx, req, annotations, pod); err != nil {
+			if code == http.StatusForbidden {
+				return admission.Denied(err.Error())
+			} else {
+				return admission.Errored(code, err)
+			}
+		}
+	} else { // use in-process evaluation
```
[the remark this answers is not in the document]
thisthat: @toddbaert @beeme1mr is this the def behavior that we want?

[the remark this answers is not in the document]
odubajDT: good point!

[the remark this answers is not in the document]
Kavindu-Dodan: Yeah, I second this suggestion. 

- if annotated with `featureflagsource`, then search for the crd of remote configurations and & inject flagd 
- if annotated with `featureflaginprocessconfiguration`, then search for the crd and inject container with in-process environment variables.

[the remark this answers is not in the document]
Kavindu-Dodan: And with the latest addition of @bacherfl 's `Flagd` crd, one could inject flagd itself as a gRPC sync source to the namespace, allowing in-process mode providers to work with flagd gRPC sync

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
Kavindu-Dodan: IMO we should make default port to `8015` so users can use flagd itself as a gRPC sync source with `Flagd` crd.

apis/core/v1beta1/common/common.go
```diff
@@ -12,6 +16,38 @@ const (
...
+	EvaluatorEnvVar             string = "EVALUATOR"
+	ImageEnvVar                 string = "IMAGE"
+	VersionEnvVar               string = "TAG"
+	ProviderArgsEnvVar          string = "PROVIDER_ARGS"
+	DefaultSyncProviderEnvVar   string = "SYNC_PROVIDER"
+	LogFormatEnvVar             string = "LOG_FORMAT"
+	ProbesEnabledVar            string = "PROBES_ENABLED"
+	DefaultEnvVarPrefix         string = "FLAGD"
+	DefaultManagementPort       int32  = 8014
+	DefaultPort                 int32  = 8013
```
Kavindu-Dodan, line 39: proposal - introrudce a new port for gRPC sync [1] named `defaultGrpcSyncPort` with value 8015 and use that for `FeatureFlagInProcessConfiguration` CRD

[1] - https://flagd.dev/reference/grpc-sync-service/

toddbaert: I think this is a good idea, maybe requirement. Good point.

Kavindu-Dodan (commented):
Overll nice work 🙌  left few suggestions and once done we can merge this :)

apis/core/v1beta1/common/common.go
odubajDT: ok, do I understand correctly that the default value of the port (currently 8013) should be changes to 8015? (only for in-process) Or do we expect an additional parameter here with 8015 as default value?

webhooks/pod_webhook.go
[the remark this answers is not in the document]
odubajDT: not sure what to take from this, is it then ok to keep this behavior do do we want to adapt it?

apis/core/v1beta1/common/common.go
toddbaert: I think basically, 8013 should be the default if we are using rpc, and 8015 should be that default if we are using in-process.

Cc @beeme1mr @Kavindu-Dodan I think we should do this for the provider defaults too outside of OFO... I think it will be a common misstep for people to change the resolver to in-process and fail to update the port to 8015. I've made this mistake myself.

webhooks/common.go
```diff
@@ -85,10 +85,37 @@ func NewFeatureFlagSourceSpec(env types.EnvConfig) *api.FeatureFlagSourceSpec {
 	}
 }
 
+func NewFeatureFlagInProcessConfigurationSpec() *api.FeatureFlagInProcessConfigurationSpec {
+	return &api.FeatureFlagInProcessConfigurationSpec{
+		Port:                  apicommon.DefaultRPCPort,
```
toddbaert, line 90:
```suggestion
		Port:                  apicommon.DefaultInProcessPort,
```

This basically doesn't allow us to correctly change the port, I think you used the wrong var

toddbaert (approved):
This works as expected with the exception of [this small bug](https://github.com/open-feature/open-feature-operator/pull/632/files#r1610638795) (tested end-to-end in playground). I will build some additional doc/tutorial later.

Approving since it's so small.

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
```diff
@@ -0,0 +1,218 @@
...
+import (
+	"fmt"
+
+	"github.com/open-feature/open-feature-operator/apis/core/v1beta1/common"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FeatureFlagInProcessConfigurationSpec defines the desired state of FeatureFlagInProcessConfiguration
+type FeatureFlagInProcessConfigurationSpec struct {
```
Kavindu-Dodan, line 28: @toddbaert related to your review comment [1]

what if we shorten the name to `InProcessConfiguration` 🤔  we don't need `FeatureFlag` prefix as this is simply configurations.  

[1] - https://github.com/open-feature/open-feature-operator/pull/632#pullrequestreview-2072356217

webhooks/common.go
odubajDT: good catch, should we allow configuration for all in-process parameters via env variables (and therefore also helm chart values)?

odubajDT: in the end I adapted it for all configurable parameters, if it's not needed, I will remove it

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
toddbaert: I think `InProcessConfiguration` is slightly better, but I don't feel very strongly about it.

@Kavindu-Dodan @beeme1mr @thisthat I think this is the only remaining concern for me (this name). If we can't think of anything better I am in favor of merging.

webhooks/pod_webhook.go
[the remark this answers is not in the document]
Kavindu-Dodan: @odubajDT I think we only need an improvement to the if-else here. `OF annotations enabled and nothing else` handling missing annotation and may be provide better error events ?

Kavindu-Dodan (approved):
LGTM :) 

Let's address these two concerns prior to merging,

1. [Conditional improvement ](https://github.com/open-feature/open-feature-operator/pull/632/files#r1604432388)
2. [Naming of the new CRD](https://github.com/open-feature/open-feature-operator/pull/632/files#r1610670361)

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
odubajDT: adapted to `InProcessConfiguration`

webhooks/pod_webhook.go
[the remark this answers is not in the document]
odubajDT: ok, adapted https://github.com/open-feature/open-feature-operator/pull/632/commits/cc098ea382edb1834dc412640916a80408182a0e

webhooks/pod_webhook.go
```diff
@@ -74,12 +76,20 @@ func (m *PodMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+			} else {
+				return admission.Errored(code, err)
+			}
+		}
+	} else if shouldUseInProcess(annotations) { // use in-process evaluation
+		if code, err := m.handleInProcessConfiguration(ctx, req, annotations, pod); err != nil {
 			return admission.Errored(code, err)
 		}
+	} else {
+		return admission.Denied("cannot mutate pods without 'featureflagsource' or 'inprocessconfiguration' annotation")
```
Kavindu-Dodan, line 92: nit - need to emphasise this is when `openfeature.dev/enabled` annotation is present.
```suggestion
		return admission.Denied("cannot mutate pods without a 'featureflagsource' or 'inprocessconfiguration' annotation as openfeature.dev/enabled annotation is set to true)
```

Kavindu-Dodan (approved):
LGTM. I like the new CRD name `InProcessConfiguration`. Left [this nit ](https://github.com/open-feature/open-feature-operator/pull/632/files#r1613622216) to improve error context. Once accepted, feel free to merge :)

apis/core/v1beta1/featureflaginprocessconfiguration_types.go
Kavindu-Dodan: Personally I like `InProcessConfiguration ` as it communicate clearly the purpose of the crd. 

@toddbaert @thisthat @beeme1mr  over to you for final approval :)

PR #633, opened 2024-05-07
feat: add `flagd` CRD with ingress support

### LINKED ISSUE #615: how to leverage OFO for workloads external to k8s , say a web frontend or something running outside k8s
I am using OFO, and it works great for backend deployments that are in my cluster, how can i use it for my frontend applications or application outside of k8s

bacherfl:
Closes #615

The PR adds a new `Flagd` CRD which can be used to deploy an instance of flagd, together with a matching service and ingress


apis/core/v1beta1/flagd_types.go
```diff
@@ -0,0 +1,124 @@
...
+
+import (
+	v1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
```
odubajDT, line 27:
```suggestion
```

config/overlays/helm/exclude-role.yaml
```diff
@@ -0,0 +1,5 @@
+$patch: delete
```
odubajDT, line 1: Why this file is needed?

config/samples/core_v1beta1_featureflagsource.yaml
```diff
@@ -1,13 +1,22 @@
 apiVersion: core.openfeature.dev/v1beta1
+kind: FeatureFlag
```
odubajDT, line 2: why it's needed to add FF sample in FFS sample?

bacherfl: that's a leftover from trying something out earlier - will remove it again

config/overlays/helm/exclude-role.yaml
bacherfl: This is because a helm value that enables this new feature has been added- since the controller for the `Flagd` resource requires additional rbac permissions (creating, deleting, updating Services and Ingresses), they will only be created if the related helm value is set to true. Therefore, a customized rbac manifest that makes use of this helm value has been added to the chart, and the default role file created by kustomized is replaced with that one

controllers/core/flagd/resource.go
```diff
@@ -0,0 +1,72 @@
...
+	newObj, err := resource.GetResource(ctx, flagd)
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("Could not create new flagd %s resource '%s/%s'", obj.GetObjectKind(), flagd.Namespace, flagd.Name))
+		return err
+	}
+
+	if exists && !resource.AreObjectsEqual(existingObj, newObj) {
+		return r.updateResource(ctx, flagd, obj, newObj)
+	}
+	return r.createResource(ctx, flagd, obj, newObj)
```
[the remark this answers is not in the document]
bacherfl: yeah good catch, that needs to be adapted of course

controllers/core/flagd/resource.go
```diff
@@ -0,0 +1,72 @@
...
+	} else if err != nil && !errors.IsNotFound(err) {
+		r.Log.Error(err, fmt.Sprintf("Failed to get flagd %s '%s/%s'", obj.GetObjectKind(), flagd.Namespace, flagd.Name))
+		return err
+	}
+
+	// check if the resource is managed by the operator.
+	// if not, do not continue to not mess with anything user generated
+	if exists && !common.IsManagedByOFO(existingObj) {
+		r.Log.Info(fmt.Sprintf("Found existing %s '%s/%s' that is not managed by OFO. Will not proceed.", obj.GetObjectKind(), flagd.Namespace, flagd.Name))
+		return fmt.Errorf("resource already exists and is not managed by OFO")
```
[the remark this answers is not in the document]
bacherfl: After some thought I decided for an error because this does prevent the Flagd resource from being reconciled by the controller, as long as there is a conflicting deployment. So passing an error here will cause the controller to cancel the current reconciliation loop and requeue for another attempt later, until the conflicting resource is no longer there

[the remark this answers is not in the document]
odubajDT: ok makes sense, thanks for explanation!

controllers/core/flagd/resource.go
```diff
@@ -0,0 +1,74 @@
...
+
+	newObj, err := resource.GetResource(ctx, flagd)
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("Could not create new flagd %s resource '%s/%s'", obj.GetObjectKind(), flagd.Namespace, flagd.Name))
+		return err
+	}
+
+	if !exists {
+		return r.createResource(ctx, flagd, obj, newObj)
+	} else if exists && !resource.AreObjectsEqual(existingObj, newObj) {
```
odubajDT, line 52: I guess we can simplify it, as this part of code we can access only if `exists == true`

```suggestion
	} else if !resource.AreObjectsEqual(existingObj, newObj) {
```

odubajDT (approved):

apis/core/v1beta1/flagd_types.go
```diff
@@ -0,0 +1,120 @@
...
+import (
+	v1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FlagdSpec defines the desired state of Flagd
+type FlagdSpec struct {
+	// Replicas defines the number of replicas to create for the service
```
toddbaert, line 27: Should we mention the default number of replicas?

apis/core/v1beta1/flagd_types.go
```diff
@@ -0,0 +1,120 @@
...
+	// +kubebuilder:validation:Enum:=ClusterIP;NodePort;LoadBalancer;ExternalName
+	ServiceType v1.ServiceType `json:"serviceType,omitempty"`
+
+	// ServiceAccountName the service account name for the flagd deployment
+	// +optional
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+
+	// OtelCollectorUri defines the OpenTelemetry collector URI to enable OpenTelemetry Tracing in flagd.
+	// +optional
+	OtelCollectorUri string `json:"otelCollectorUri"`
```
toddbaert, line 45: This already [exists](https://github.com/open-feature/open-feature-operator/blob/b58261976d03059fe7b214244eb8087519cbbc7c/apis/core/v1beta1/featureflagsource_types.go#L105-L107) on the `FeatureFlagSource` spec. Do the do the same thing? I'm a bit confused why it's on both :sweat_smile:

chart/open-feature-operator/README.md
```diff
@@ -126,24 +126,37 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `controllerManager.manager.resources.requests.memory`                     | Sets memory resource requests for operator.                                                                                      | `64Mi`                                       |
+| `controllerManager.replicas`                                              | Sets number of replicas of the OpenFeature operator pod.                                                                         | `1`                                          |
+| `managerConfig.flagsValidatonEnabled`                                     | Enables the validating webhook for FeatureFlag CR.                                                                               | `true`                                       |
+| `managerConfig.flagdResourceEnabled`                                      | Enables the controller for the Flagd CR and adds the required permissions to create/update/delete Service and Ingress resources. | `true`                                       |
```
toddbaert, line 159:
```suggestion
| `managerConfig.flagdResourceEnabled`                                      | Enables the controller for the Flagd CR and adds the required permissions to automatically manage the exposure of flagd via Service and Ingress resources. | `true`                                       |
```

Maybe you or somebody else can word this better, but I think a "why" was missing here, and I've tried to capture that.

chart/open-feature-operator/values.yaml
```diff
@@ -92,6 +110,8 @@ controllerManager:
 managerConfig:
   ## @param managerConfig.flagsValidatonEnabled Enables the validating webhook for FeatureFlag CR.
   flagsValidatonEnabled: "true"
+  ## @param managerConfig.flagdResourceEnabled Enables the controller for the Flagd CR and adds the required permissions to create/update/delete Service and Ingress resources.
```
toddbaert, line 113: I guess this is the real source of truth for my comment [here](https://github.com/open-feature/open-feature-operator/pull/633/files?diff=split&w=1#r1598677911).

test/e2e/kuttl/flagd-with-ingress-custom-paths/00-assert.yaml
```diff
@@ -0,0 +1,120 @@
+apiVersion: apps/v1
```
toddbaert, line 1: Tests make perfect sense. :+1:

toddbaert (commented):
This looks great to me, but I have some confusion with this, and I think we need a very small addition covering [this](https://github.com/open-feature/open-feature-operator/pull/633/files#r1598662844) in [docs](https://github.com/open-feature/open-feature-operator/tree/main/docs).

common/flagdproxy/flagdproxy.go
```diff
@@ -121,7 +120,7 @@ func (f *FlagdProxyHandler) newFlagdProxyServiceManifest(ownerReference *metav1.
 		Spec: corev1.ServiceSpec{
 			Selector: map[string]string{
 				"app.kubernetes.io/name":       FlagdProxyDeploymentName,
-				"app.kubernetes.io/managed-by": ManagedByAnnotationValue,
+				"app.kubernetes.io/managed-by": common.ManagedByAnnotationValue,
```
Kavindu-Dodan, line 123: nit - can use `ManagedByAnnotationKey` defined in common

go.mod
```diff
@@ -79,4 +79,7 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-replace golang.org/x/net => golang.org/x/net v0.24.0
+replace (
+	github.com/open-feature/open-feature-operator/apis => github.com/bacherfl/open-feature-operator/apis v0.0.0-20240508085321-08cc258ed722
```
Kavindu-Dodan, line 83: Since we merged https://github.com/open-feature/open-feature-operator/pull/635 , you should be able to test things easily. Anyway please check the replacement :)

apis/core/v1beta1/flagd_types.go
Kavindu-Dodan: I think this is not relevant to the service definition. Same goes for the `FeatureFlagSource` 🤔

common/types/envconfig.go
```diff
@@ -22,4 +31,5 @@ type EnvConfig struct {
 	SidecarSyncProvider  string `envconfig:"SIDECAR_SYNC_PROVIDER" default:"kubernetes"`
 	SidecarLogFormat     string `envconfig:"SIDECAR_LOG_FORMAT" default:"json"`
 	SidecarProbesEnabled bool   `envconfig:"SIDECAR_PROBES_ENABLED" default:"true"`
+	FlagdResourceEnabled bool   `envconfig:"FLAGD_RESOURCE_ENABLED" default:"true"`
```
Kavindu-Dodan, line 34: nit - we should document this env flag and guide on how to use it

controllers/core/flagd/resource.go
```diff
@@ -0,0 +1,74 @@
+package flagd
```
Kavindu-Dodan, line 1: +1 for having a common reconciler. can we rename the file to something like `resource_reconciler` so the idea is clear ?

controllers/core/flagd/controller.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	if err := r.ResourceReconciler.Reconcile(
+		ctx,
+		flagd,
+		&v1.Service{},
+		r.FlagdService,
+	); err != nil {
+		return ctrl.Result{}, err
+	}
```
Kavindu-Dodan, line 98: Should we perform this manual reconcile if `flagd.Spec.Ingress.Enabled` is disabled ? I assume the service and the ingress are related and only aplicable in this specific mode of operation ? 🤔

apis/core/v1beta1/flagd_types.go
```diff
@@ -0,0 +1,120 @@
...
+package v1beta1
+
+import (
+	v1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// FlagdSpec defines the desired state of Flagd
+type FlagdSpec struct {
```
Kavindu-Dodan, line 26: I know naming is hard ;) but should we emphasise this a side-car in this naming - `FlagdSidecarSpec` ?

controllers/core/flagd/resources/deployment.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := r.Client.Get(ctx, client.ObjectKey{
+		Namespace: flagd.Namespace,
+		Name:      flagd.Spec.FeatureFlagSource,
+	}, featureFlagSource); err != nil {
+		return nil, fmt.Errorf("could not look up feature flag source for flagd: %w", err)
+	}
+
+	err := r.FlagdInjector.InjectFlagd(ctx, &deployment.ObjectMeta, &deployment.Spec.Template.Spec, &featureFlagSource.Spec)
+	if err != nil {
+		return nil, fmt.Errorf("could not inject flagd container into deployment: %w", err)
```
Kavindu-Dodan, line 90: do we need to inject flagd in this reconcilation loop 🤔

apis/core/v1beta1/flagd_types.go
Kavindu-Dodan: Besides, for my understanding, is this created per namespace where we expect special flagd deployment manipulation with ingress and service ? :)

go.mod
bacherfl: indeed, now that #635 is merged this replace is not needed anymore :)

controllers/core/flagd/controller.go
bacherfl: I would argue yes, as you may not want to expose the service via an Ingress, but still want to use the new CRD to have a flagd deployment + service that you can access within the cluster

apis/core/v1beta1/flagd_types.go
bacherfl: I would keep it as `FlagdSpec` in this case, as this contains all the configuration options for a standalone flagd deployment, service and ingress - `FlagdSidecarSpec` would suggest that the options available here are only for a single flagd container in my opinion

controllers/core/flagd/resources/deployment.go
bacherfl: yes, i thought it makes sense to reuse the existing `FlagdInjector` to insert the flagd container in the deployment to be created here, as this component also sets up the required rbac permissions for the service account of the deployment to be able to read `FeatureFlag` CRs in the cluster

apis/core/v1beta1/flagd_types.go
bacherfl: and yes, the Flagd CRD is namespaced, so the deployment/service/ingress will be created in the same namespace as the `Flagd` resource :)

chart/open-feature-operator/README.md
bacherfl: thanks for the suggestion, this makes things a bit clearer

apis/core/v1beta1/flagd_types.go
bacherfl: @toddbaert you are right, having this here is not required, as it's already in the FeatureFlagSource - will remove this

common/types/envconfig.go
bacherfl: this is documented in the related helm value here: https://github.com/open-feature/open-feature-operator/pull/633/files#diff-c1955654bbc791643783433fa9cc013ea352647c08712007bb5af430ca398b7bR113 

The reason why this flag has been added is because handling this new CRD requires additional permissions for the operator, which may not be something users would like to grant if they do not intend to make use of this new feature

Kavindu-Dodan: Okay, yes can disable this for permission reasons. That's a valid point 👍

controllers/core/flagd/controller.go
Kavindu-Dodan: @bacherfl ahh nice, makes sense 👍

apis/core/v1beta1/flagd_types.go
Kavindu-Dodan: Another option is `FlagdCustomizationSpec` ;)  feel free to ignore.

controllers/core/flagd/resources/deployment.go
Kavindu-Dodan: for my understanding, we already do this at pod webhook, so here it's an override based on new crd definition isn't it ?

Kavindu-Dodan (approved):
Thank you @bacherfl for this feature and clarifications 🙌

toddbaert (approved):

toddbaert:
Thanks @bacherfl

Issue #542, opened 2023-11-14
[feature] Support in-process evaluator in OFO


toddbaert:
We now have JS, Java, and Go [in-process flagd providers](https://flagd.dev/architecture/#in-process-evaluation). These providers connect directly to a `sync.proto`-compliant gRPC server (such as [flagd-proxy](https://github.com/open-feature/flagd/pkgs/container/flagd-proxy)). OFO should be enhanced to support the use of these providers. This would required:

- ability to specify this in APIs
- OFO sets the corresponding env-var which configures the provider to use in-process evaluation (see [flagd-injector](https://github.com/open-feature/open-feature-operator/blob/a97d336468d5a9b50662f4979784c8388ec10ec1/controllers/common/flagd-injector.go#L72C3-L72C7)), `FLAGD_SOURCE_RESOLVER`
- flagd _is not deployed as a sidecar to the associated workload_, since it's not needed

odubajDT:
After some discussions, it would be beneficial to do a refactoring of the existing `FeatureFlagSource` CRD structure to be able to distinguish which parameters are set for `RPC` or `in-process` evaluation. We agreed to draft a PoC with the potential solution (with bumping the API version to `v1beta2`) also containing the conversion webhooks, so users running v1beta1 versions won't need to adapt their resources in case they want to use only RPC evaluation.

odubajDT:
Implemented a [PoC](https://github.com/open-feature/open-feature-operator/pull/628) supporting the in-process evaluation and additionally refactoring the actual API. During the implementation, I got the feeling that it would be possible to move the in-process evaluation parameters into a new separate CRD to decrease the complexity of FeatureFlagSource CRD and therefore reducing the need to introduce conversion webhooks. Also it can potentially bring the advantage of easier testability of the code sa well as decresing the complexity of the mutating webhook logic (especially the part handling the decision if the flagd sidecar injection should happen). Open for any opinions!

@toddbaert

bacherfl:
In general I also have the impression that the current `FeatureFlagSource` is handling too many different responsibilities, so if we can extract something into a separate CRD i think that would be beneficial

Kavindu-Dodan:
@odubajDT what you are proposing is to keep `FeatureFlagSource` untouched and introduce a new CRD dedicated for in-process mode isn't it ? 

I like the idea. We should also consider the ability to utilize flagd's grpc streaming capability [^1] which was added recently (@toddbaert we could refine the issue on this).

[...]

PR #634, opened 2024-05-08
fix: inject env variables to all pod containers

odubajDT:
Part-of: #542 

## Changes

 - fixed [small issue](https://github.com/open-feature/open-feature-operator/pull/634/files#diff-edf09fdaf271cf47315a206784c3406e9a895fa16bae1ce11e5c0f4ef79d93d1R73) in flaginjector about adding env variables to all containers (not only flagd sidecar)
 - adapt flagdinjector tests to test the changed behavior + simplify them
 - polish webhook for future extension with in-process evaluation

### DIFF
--- a/common/flagdinjector/flagdinjector.go
+++ b/common/flagdinjector/flagdinjector.go
@@ -72,8 +72,7 @@ func (fi *FlagdContainerInjector) InjectFlagd(
 
 	flagdContainer.Env = append(flagdContainer.Env, flagSourceConfig.ToEnvVars()...)
 	for i := 0; i < len(podSpec.Containers); i++ {
-		cntr := podSpec.Containers[i]
-		cntr.Env = append(cntr.Env, flagdContainer.Env...)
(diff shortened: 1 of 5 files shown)


common/flagdinjector/flagdinjector.go
```diff
@@ -29,8 +29,7 @@ const (
 type IFlagdContainerInjector interface {
 	InjectFlagd(
 		ctx context.Context,
-		objectMeta *metav1.ObjectMeta,
-		podSpec *corev1.PodSpec,
+		pod *corev1.Pod,
```
bacherfl, line 32: is there a chance we can keep using `PodSpec` here? Otherwise the `Flagd` CRD controller which is also using this interface for inserting a flagd container into the PodSpec of a `Deployment` will not be able to use this easily anymore (at least it will be a bit awkward to use in that case)

odubajDT: hmm yes I see, I guess I will revert the change

toddbaert (approved):

bacherfl (approved):

PR #635, opened 2024-05-08
chore: use workspaces to make api changes easier

bacherfl:
This PR introduces the use of go workspaces. Before this change, anything that involves a change or addition in the `apis` submodule required this to be split into multiple PRs (one for the api change, then a follow up to adapt the code in the controller/webhook). A similar approach is also used in https://github.com/open-feature/flagd, so I think we could also make use of that here

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -44,6 +44,9 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Workspace Init
+        run: make workspace-init
+
       - name: Unit Test
         run: make unit-test
 
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -122,6 +122,7 @@ jobs:
           with:
             go-version: ${{ env.DEFAULT_GO_VERSION }}
         - run: |
+            make workspace-init
             go mod tidy
             make controller-gen
             IMG=ghcr.io/open-feature/open-feature-operator:${{ needs.release-please.outputs.release_tag_name }} make helm-package
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 
 # Copy the go source
-COPY apis/ apis/
 COPY main.go main.go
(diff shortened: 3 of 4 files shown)


odubajDT (approved):

bacherfl:
> Great idea!
> 
> One question: Is there any CI (particularly the [release CI](https://github.com/open-feature/open-feature-operator/blob/main/.github/workflows/release-please.yml)) that needs to be adapted for this to work? Do those flows need to init the workspace?

Looking at the CI I believe that the workspace might need to be initialised here: https://github.com/open-feature/open-feature-operator/blob/2c394282592bf9f6626c80bdeea2e5e20cabd274/.github/workflows/release-please.yml#L125 

although i'm not sure if the `go mod tidy` and `make controller-gen` is actually needed - do you have some background information on why this was added here?

toddbaert:
> > Great idea!
> > One question: Is there any CI (particularly the [release CI](https://github.com/open-feature/open-feature-operator/blob/main/.github/workflows/release-please.yml)) that needs to be adapted for this to work? Do those flows need to init the workspace?
> 
> Looking at the CI I believe that the workspace might need to be initialised here:
> 
> https://github.com/open-feature/open-feature-operator/blob/2c394282592bf9f6626c80bdeea2e5e20cabd274/.github/workflows/release-please.yml#L125
> 
> although i'm not sure if the `go mod tidy` and `make controller-gen` is actually needed - do you have some background information on why this was added here?
> 
> Edit: also I noticed the workspace init was missing before the test execution, so i have added it here as well

Based on blame, it's been this way for 2 years: https://github.com/open-feature/open-feature-operator/commit/d4abda119e4a7c2dab7a2e0d335d44b1df07ec62

It seems like this is(was?) necessary before running `... make helm-package` and `... make release-manifests` on the next lines. I don't see any Go dependency management or anything like that in the Makefile, so I guess that's why it's needed here? Does that make sense?

Kavindu-Dodan (approved):

toddbaert:
Merging this now... we're releasing today so we'll know if it works :sweat_smile:

toddbaert:
Worked fine with the workspace init. I think we could move this into the makefile to make things more self-contained, but at least everything works as is.

Issue #630, opened 2024-05-03
sidecarConfiguration.resources in helm chart is ignored


mng1dev:
Hello, 

I am trying to use the operator and I am installing it via helm chart.

I am passing the following `values.yaml` file, but it doesn't seem to affect the default properties of the `flagd` sidecar.

Am I doing anything wrong?

```yaml
sidecarConfiguration:
  resources:
    requests:
      cpu: 10m
      memory: 16M
    limits:
      cpu: 50m
      memory: 32M
```

toddbaert:
```
USER-SUPPLIED VALUES:
sidecarConfiguration:
  port: 8078
  resources:
    limits:
      memory: 72M

COMPUTED VALUES:
controllerManager:
...
  resources:
    limits:
      memory: 72M
...
```

The helm template seems to be working, but the flagd sidecar doesn't seem to use these resources. The other customizations from the `sidecarConfiguration` seem to work (port, etc), so I think there's a bug here. I'm still investigating.

toddbaert:
Thanks again @mng1dev . This is fixed with https://github.com/open-feature/open-feature-operator/pull/639 and will be release soon.

PR #639, opened 2024-05-14
fix: helm sidecar resources not applied

toddbaert:
We weren't correctly overlaying these in helm. I've added them similarly to others, using the same defaults from from [here](https://github.com/open-feature/open-feature-operator/blob/65e20cf72b3e1c90e3c3a6ab714fd82c2189cd33/main.go#L59-L62).

Fixes: https://github.com/open-feature/open-feature-operator/issues/630

Manually tested and works as expected as well.


chart/open-feature-operator/README.md
```diff
@@ -99,22 +99,25 @@ The command removes all the Kubernetes components associated with the chart and
...
-| `sidecarConfiguration.otelCollectorUri`    | Otel exporter uri.                                                                                                                                                                                                                                          | `""`                         |
-| `sidecarConfiguration.resources`           | Override resources of the flagd sidecar.                                                                                                                                                                                                                    | `{}`                         |
+| Name                                             | Description                                                                                                                                                                                                                                                 | Value                        |
```
toddbaert, line 102: autogenerated.

Kavindu-Dodan (approved):
LGTM

Basically we have two options to configure injected flagd's resource requests,

1. From operator Helm (once this PR fix the bug)
2. Through `FeatureFlagSource` resources where we can override this per deployment

I think we need to document these concepts :)

beeme1mr (approved):

chart/open-feature-operator/values.yaml
```diff
@@ -33,8 +33,17 @@ sidecarConfiguration:
   debugLogging: false
   ## @param sidecarConfiguration.otelCollectorUri Otel exporter uri.
   otelCollectorUri: ""
-  ## @param sidecarConfiguration.resources Override resources of the flagd sidecar.
-  resources: { }
+  resources:
+    limits:
```
thisthat, line 37: This would remove support for Resource Claims :( 
https://kubernetes.io/docs/concepts/scheduling-eviction/dynamic-resource-allocation/

See the types here: https://github.com/kubernetes/api/blob/master/core/v1/types.go#L2604C11-L2604C24

thisthat (approved):

odubajDT (approved):

PR #647, opened 2024-05-29
fix: adapt rolebinding to modified manager role

### LINKED ISSUE #645: [BUG] Helm deploy broken
OFO v0.5.6 never starts up when installed with Helm. 

With OFO version `0.5.6`, from the `manager` pod logs, I get:

```
failed to list *v1 │
│ .Deployment: deployments.apps is forbidden: User "system:serviceaccount:open-feature-operator-system:open-feature-operator-contro │
│ ller-manager" cannot list resource "deployments" in API group "apps" at the cluster scope: RBAC: clusterrole.rbac.authorization.k │
│ 8s.io "manager-role" not found
```

The pod fails to become ready. This is related to this newly added helm overlay [here](https://github.com/open-feature/open-feature-operator/pull/633/files#diff-ee17a7151ecd6906c72b490a136d7e6972c12733882bde3856a1804e277cecca). **_If I remove this overlay, things work as expected._** Keeping it results in the role not existing in the cluster.

Possibly related, I noticed this delta in the output charts (comparing `v0.5.5` to `v0.5.6`): 

image

bacherfl:
Closes #645 

This PR solves the issue by also adapting the ClusterRoleBinding using the recently modified ClusterRole. The reason this was failing is because due to the `delete` overlay for the `manager-role`, the `open-feature-operator` prefix was not added to the [ClusterRoleBinding](https://github.com/open-feature/open-feature-operator/blob/45651aeac19d1276ae4d4d73820a9af93ef67a90/config/rbac/role_binding.yaml#L8) anymore, which resulted in the ClusterRoleBinding referring to the wrong ClusterRole (`manager-role` instead of `open-feature-operator-manager-role`)


beeme1mr (approved):

toddbaert:
Rendered chart looks right when tested locally. Merging.

toddbaert (approved):

PR #648, opened 2024-05-29
fix: include parameters with default values to envVars

### LINKED ISSUE #646: [BUG] InProcessConfiguration not injecting env-vars if set to default
If I set an InProcessConfiguration like this:

```yaml
apiVersion: core.openfeature.dev/v1beta1
kind: InProcessConfiguration
metadata:
    name: in-process-config
spec:
    host: flagd-standalone
    port: 8015
```
the resulting container looks like this (note `FLAGD_PORT` is missing):

image

If I change the port to something other than `8015` like this:

```yaml
apiVersion: core.openfeature.dev/v1beta1
kind: InProcessConfiguration
metadata:
    name: in-process-config
spec:
    host: flagd-standalone
    port: 8033
```

Then the var is set (note `FLAGD_PORT` is now set to `8033` as expected): 

image

odubajDT:
## This PR

- includes parameters with default values in the injected envVar containers (for `FeatureFlagSource` and `InProcessConfiguration`)

### Related Issues

Fixes #646


toddbaert (approved):

toddbaert:
Thanks!

Issue #651, opened 2024-05-29
Helm installation broken


toddbaert:
Helm installation is not working due to the recently added overlays removing rules from the `open-feature-operator-manager-role`. The operator is now not able to read openfeature.dev/* custom resources when installed with Helm:

```
failed to list *v │
│ 1beta1.InProcessConfiguration: inprocessconfigurations.core.openfeature.dev is forbidden: User "system:serviceaccount:open-featur │
│ e-operator-system:open-feature-operator-controller-manager" cannot list resource "inprocessconfigurations" in API group "core.ope │
│ nfeature.dev" at the cluster scope
```

![image](https://github.com/open-feature/open-feature-operator/assets/25272906/a8d2742b-a8ca-4901-a813-7c7a49effc31)

toddbaert:
@thisthat @bacherfl When I checked the output charts in https://github.com/open-feature/open-feature-operator/pull/647, the role-binding looked fine, but I never checked the role which is now more messed up :sweat_smile: 

I will see how far I can get with this, but I wonder if we should change [the patch to not be a delete](https://github.com/open-feature/open-feature-operator/blob/1ea457fdcb389893daa7165cc8b99165564b0ba6/config/overlays/helm/exclude-role.yaml#L1) and instead add the additional rules we need for ingress creation?

toddbaert:
I started down the path of maintaining 2 roles and 2 rolebindings, one that controls the extra stuff we need when `managerConfig.flagdResourceEnabled` is enabled, but after running through some proposed tutorials and dealing with this additional complexity, I honestly think the entire idea of `managerConfig.flagdResourceEnabled` is not worth it. My reasoning is that it's easy to restrict the creation of flagd CR instances entirely (or instances with `ingresses`) through many different tools and policies, since these are not created automatically by annotations. I just don't think we need to control them at the OFO install level.

Sorry @bacherfl for the wasted time here, since this was initially my idea. :pensive:

PR #652, opened 2024-05-29
fix!: remove flagdResourceEnabled

toddbaert:
For https://github.com/open-feature/open-feature-operator/issues/651, I started down the path of maintaining 2 roles and 2 rolebindings, one that controls the extra stuff we need when `managerConfig.flagdResourceEnabled` is enabled, but after running through some proposed tutorials and dealing with this additional complexity, I honestly think the entire idea of `managerConfig.flagdResourceEnabled` is not worth it. My reasoning is that it's easy to restrict the creation of flagd CR instances entirely (or instances with ingresses) through many different tools and policies, since these are not created automatically by annotations like some of our other resources. I just don't think we need to control them at the OFO install level.

Sorry @bacherfl for the wasted time here, since this was initially my idea. 😔

Fixes: https://github.com/open-feature/open-feature-operator/issues/651

I tested this locally and it works as expected with Helm.


config/manager/kustomization.yaml
```diff
@@ -10,5 +10,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: controller
+  newName: ghcr.io/openfeature/operator
```
toddbaert, line 13:
```suggestion
  newName: controller
```

Kavindu-Dodan:
I am fine with removing the `flagdResourceEnabled ` option and getting the helm release to a stable state.

Issue #665, opened 2024-06-04
Kustomize/Helm Cleanup / Support Multiple ImagePullSecrets


toddbaert:
As discussed in https://github.com/open-feature/open-feature-operator/pull/655, the fact we use both Kustomize and Helm causes us some pain and limitations. We should see if we can completely remove kustomize in favor of pure Helm, and as a first-fruit of this effort, allow multiple `ImagePullSecrets` to be defined (instead of the current limitation to one).

> One request I'll have is to add supporting the array of imagePullSecrets to the backlog for OpenFeature. Inevitably the build process will change and evolve, and when it does, this use-case will just become part of it.

_Originally posted by @cpitstick-latai in https://github.com/open-feature/open-feature-operator/issues/655#issuecomment-2146144710_

toddbaert:
@thisthat and I spend some time looking into this. I actually don't think it's reasonably possible right now unless we want to manually maintain a chart (which I'm really reluctant to do).

The main issue is that kubebuilder is the thing that generates our kustomize layer from go-code (which is really useful). There's no way to have kubebuilder generate helm YET, but there is a [feature/plugin in discussion](https://github.com/kubernetes-sigs/kubebuilder/pull/3632).

PR #666, opened 2024-06-05
fix: handle multiple imagePullSecrets

toddbaert:
The previous iteration of this was ALMOST working, but:

- it didn't support multiple secrets
- it broke if you did a helm upgrade without a imagePullSecret, forcing the user to use `--force`.

I've added a sort of hacky "post-processor" script which allows us to add helm control flow where we otherwise couldn't. After talking with @thisthat I think this is the only solution until kustomize supports helm output.

Again, I tested locally, and ImagePullSecrets (even more than one) are added to both the manager deployment and the deployments we create at runtime.

Fixes: https://github.com/open-feature/open-feature-operator/issues/665

### DIFF
--- a/config/overlays/helm/manager.yaml
+++ b/config/overlays/helm/manager.yaml
@@ -92,7 +92,7 @@ spec:
             - --sidecar-ram-limit={{ .Values.sidecarConfiguration.resources.limits.memory }}
             - --sidecar-cpu-request={{ .Values.sidecarConfiguration.resources.requests.cpu }}
             - --sidecar-ram-request={{ .Values.sidecarConfiguration.resources.requests.memory }}
-            - --image-pull-secret={{ range .Values.imagePullSecrets }}{{ .name }},{{- end }}
+            - --image-pull-secrets={{ range .Values.imagePullSecrets }}{{ .name }},{{- end }}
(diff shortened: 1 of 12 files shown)


config/overlays/helm/manager.yaml
```diff
@@ -92,7 +92,7 @@ spec:
             - --sidecar-ram-limit={{ .Values.sidecarConfiguration.resources.limits.memory }}
             - --sidecar-cpu-request={{ .Values.sidecarConfiguration.resources.requests.cpu }}
             - --sidecar-ram-request={{ .Values.sidecarConfiguration.resources.requests.memory }}
-            - --image-pull-secret={{ .Values.imagePullSecret }}
+            - --image-pull-secret={{ range .Values.imagePullSecrets }}{{ .name }},{{- end }}
```
toddbaert, line 95: spread the array of secret objects into our arg, delimited by `,`

Kavindu-Dodan (approved):
I think this is a good solution to overcome the limitation 👍

thisthat (approved):
Unfortunately, I believe this is the best solution we could use at the moment until there is a better support for Helm control flow strings in Kustomize :(

PR #679, opened 2024-06-17
feat!: split bind address to manage host and port separately

cpitstick-latai:
Adds webhook port and bind port for metrics server to the Deployment for the main operator. Also allows customization of the metrics bind port for the operator container. This is to prevent port conflicts when `hostNetwork: true` is set (https://github.com/open-feature/open-feature-operator/pull/680)

## This PR
The bindPort for the metrics server was not customizable for the open-feature-operator. Enabling this to be configured. Also explicitly specify both the metrics server port and the webhook port in the deployment.

### Notes
This is necessary for enabling `hostNetwork: true` which is in [this PR](https://github.com/open-feature/open-feature-operator/pull/680)

### How to test
On an EKS cluster where the CNI has been replaced with Calico, and combined with `hostNetwork: true`


chart/open-feature-operator/values.yaml
```diff
@@ -148,8 +148,10 @@ managerConfig:
       ## @param managerConfig.controllerManagerConfigYaml.health.healthProbeBindAddress Sets the bind address for health probes.
       healthProbeBindAddress: :8081
     metrics:
-      ## @param managerConfig.controllerManagerConfigYaml.metrics.bindAddress Sets the bind address for metrics.
-      bindAddress: 127.0.0.1:8080
+      ## @param managerConfig.controllerManagerConfigYaml.metrics.bindAddress Sets the bind address for metrics (combined with bindPort).
+      bindAddress: 127.0.0.1
+      ## @param managerConfig.controllerManagerConfigYaml.metrics.bindPort Sets the bind port for metrics.
+      bindPort: :8080
```
toddbaert, line 154: Things are a bit inconsistent here. We have a few options that are just ints, and a few that look like `:{int}`.

Should we make this consistent? @thisthat @bacherfl @odubajDT What do you guys think?

bacherfl: I think in this case it might suffice to just have the `bindAddress` parameter as a string and have this set to `:8080` by default.

I think the reason we have those options that are int and ones that look like `:{int}` is because those parameters are used as `string` (e.g. `:8080`) and `int` for the related go types [here](https://github.com/open-feature/open-feature-operator/blob/339e5c8cf49c54660ed3ee70fe988345b55de999/main.go#L135) and [here](https://github.com/open-feature/open-feature-operator/blob/339e5c8cf49c54660ed3ee70fe988345b55de999/main.go#L146) (the webhook server port seems to be hard coded as well btw)

So i think we can probably keep it this way

cpitstick-latai: I'm unclear on what changes you want made here?

Re-write so that bindAddress is simply `:8080` and remove the `127.0.0.1`? Wouldn't that hurt flexibility for if a user really wanted another address?

The webhook server port can be addressed in a separate PR.

bacherfl: I would suggest to just use `bindAddress` and set that to `:8080` by default. This makes it a bit clearer that this value directly relates to the `--metrics-bind-address` flag of the operator. The value provided here does not need to be only a port, but can contain the address as well (see https://pkg.go.dev/net#Listen), so this should not reduce flexibility.
regarding the webhook server port, I agree that - if necessary - this can be done in a separate PR

cpitstick-latai: So I've worked on this a little further and concluded:
- Actually we should do webhook port with this one as both ports are relevant and could cause a port conflict when `hostNetwork` is true. I've added it to the deployment.
- I've also added the bindPort to the deployment as well, but this means we can't have a `:` in front of the port number and have to introduce that in the Helm templating, as it's required to be a number (I tested it!).

toddbaert:
My only concern is that things still aren't quite consistent. For example, `managerConfig.controllerManagerConfigYaml.health.healthProbeBindAddress` includes the `:` and doesn't work the same as these new params (separate host/port). However, it's the case that things already weren't quite consistent here, and I'm not sure if it's worth addressing in this PR or not.

I'm approving, but I'd like to hear from others. cc @bacherfl @thisthat @odubajDT

toddbaert (approved):

cpitstick-latai:
@toddbaert A dev anti-pattern is to hold a PR hostage by demanding that changes that are not in the PR's scope be made to fix other technical debt. I saw this early in my career, and have learned to aggressive fight it.

Fixing the consistency here can be the NEXT PR. But I'm not sure I'm the right person to do this fix, as your team seems to have a particular and peculiar view of what you desire.

beeme1mr (approved):

bacherfl (approved):
lgtm

toddbaert:
> @toddbaert A dev anti-pattern is to hold a PR hostage by demanding that changes that are not in the PR's scope be made to fix other technical debt. I saw this early in my career, and have learned to aggressive fight it.
> 
> Fixing the consistency here can be the NEXT PR. But I'm not sure I'm the right person to do this fix, as your team seems to have a particular and peculiar view of what you desire.

Agreed, we just need 2 approvals and I wanted to point out the implicit choice. I'm fine to merge as is.

PR #680, opened 2024-06-17
feat: Add hostNetwork flag.

cpitstick-latai:
## This PR
Add `hostNetwork` and `dnsPolicy` settings to enable compatibility with EKS clusters that have a custom CNI. Without these, the MutatingWebhook that adds the flagd sidecar to pods will not function.

### Related Issues
- This provided inspiration for this fix: https://github.com/elastic/cloud-on-k8s/issues/1369
- Another issue with the same problem/fix: https://stackoverflow.com/questions/59801157/aws-eks-k8s-cluster-with-weave-cni-plugin-cannot-have-custom-metrics
- The HashiCorp vault sidecar injector explicitly notes this in its Helm chart values as well: https://github.com/hashicorp/vault-helm/blob/834cd7f85a48179bef83ce48f82aec2128bc6020/values.yaml#L328

### Notes
This relies on `.github/scripts/strip-kustomize-helm.sh` because boolean values have to be true/false without quotes, so to set `hostNetwork` to true in the helm chart, we have to strip out the quotes that are used to allow kustomize to ignore helm chart directives.

### Follow-up Tasks
- The ongoing issues related to the kustomize build process and the conflict with Helm continue to plague this type of system.

### How to test
Test the operator with a test workload that uses the flagd sidecar on an EKS cluster with a custom CNI (Calico is most common).

### DIFF
--- a/chart/open-feature-operator/README.md
+++ b/chart/open-feature-operator/README.md
@@ -158,22 +158,24 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Operator resource configuration
 
-| Name                                                                      | Description                                              | Value                                        |
-| ------------------------------------------------------------------------- | -------------------------------------------------------- | -------------------------------------------- |
-| `controllerManager.kubeRbacProxy.image.repository`                        | Sets the image for the kube-rbac-proxy.                  | `gcr.io/kubebuilder/kube-rbac-proxy`         |
-| `controllerManager.kubeRbacProxy.image.tag`                               | Sets the version tag for the kube-rbac-proxy.            | `v0.14.1`                                    |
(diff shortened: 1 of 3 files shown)


toddbaert:
This seems right to me. The CNCF mandates you sign-off your commit though, so we implement some checks for that.

You can do `git commit --amend --signoff` and force push.

cpitstick-latai:
I did a sign off. However, it looks like the actions workflow experienced a transient failure!

cpitstick-latai:
This should be ready to go now!

toddbaert (approved):

toddbaert:
Tested locally, looks good. I'll give 24h for others to review, but I think this is small enough to merge with 1 approval.

thisthat (approved):

PR #681, opened 2024-06-18
feat: Add labels and annotations to pods.

cpitstick-latai:
Currently, there is only one deployment, so avoid premature optimization by making these global (and follow the same pattern for imagePullSecrets)

Requires more aggressive usage of the `strip-kustomize-helm.sh` script because we have to preserve existing annotations and labels from the config/manager directory.

### Notes
I chose to use `maps.Copy` which requires an upgrade in some of the golang dependencies. This function was made standard in Golang 1.21, which is the version the operator currently uses, so I don't foresee issues here.

### Testing

There are 3 use-cases that matter here:
- Sidecar injection
- FlagD standalone
- FlagD proxy.

Label injection should happen only on the last two. Though I haven't tested those. I verified that this works on an EKS 1.27 cluster with a Calico CNI and verified sidecar injection doesn't break.


.github/scripts/strip-kustomize-helm.sh
```diff
@@ -1,18 +1,26 @@
...
+    "s/___space___/ /g"
+    "s/\"___//g"
+    "s/___\"//g"
+    "/___delete_me___/d"
+    "s/___//g"
+)
+find $CHARTS_DIR -name "*.yaml" | while read file; do
+    for expr in "${sed_expressions[@]}"; do
+        sed -i "$expr" "$file"
+    done
```
toddbaert, line 23: :sunglasses: 

Now we own this together. :imp:

toddbaert:
I'll give this and your other PRs a test tomorrow.

.github/scripts/strip-kustomize-helm.sh
cpitstick-latai: Nothing is more permanent than a temporary hack :)

toddbaert:
Hey @cpitstick-latai this is a great start! I was able to easily get the annotations working locally after running `make helm-release` and setting some in my `values.yaml`. I think I see 2 issues:

- NOT passing any values seems to break the install. It's not immediately obvious to me why, since your defaults seem correct. In any case, I see: 

```sh
helm upgrade --install open-feature-operator  /home/todd/git/open-feature-operator/charts/open-feature-operator-v0.6.1.tgz  
Release "open-feature-operator" does not exist. Installing it now.
Error: YAML parse error on open-feature-operator/templates/apps_v1_deploymen
```

- This PR doesn't deal with the runtime pods created by the operator. I think the annotations/labels should be added there as well, similar to how we added imagePullSecret there. See https://github.com/open-feature/open-feature-operator/pull/655 and https://github.com/open-feature/open-feature-operator/pull/666

thisthat:
> Hey @cpitstick-latai this is a great start! I was able to easily get the annotations/labels working locally after running `make helm-release` and setting some in my `values.yaml`. I think I see 2 issues:
> 
> * NOT passing any values seems to break the install. It's not immediately obvious to me why, since your defaults seem correct. In any case, I see:
> 
> ```shell
> helm upgrade --install open-feature-operator  /home/todd/git/open-feature-operator/charts/open-feature-operator-v0.6.1.tgz  
> Release "open-feature-operator" does not exist. Installing it now.
> Error: YAML parse error on open-feature-operator/templates/apps_v1_deployment_open-feature-operator-controller-manager.yaml: error converting YAML to JSON: yaml: line 16: did not find expected key
> ```
> 
> * This PR doesn't deal with the runtime pods created by the operator. I think the annotations/labels should be added there as well, similar to how we added imagePullSecret there. See [feat: add image pull secrets #655](https://github.com/open-feature/open-feature-operator/pull/655) and [fix: handle multiple imagePullSecrets #666](https://github.com/open-feature/open-feature-operator/pull/666)

This sounds that we need to implement a CI action for testing the Helm installation in the e2e tests 🤔

cpitstick-latai:
I think I fixed the issue where empty dictionaries `{}` break.

Thinking about how to implement the runtime pod issue.

cpitstick-latai:
@toddbaert I believe I've resolved all the issues you pointed out. I've largely followed your blueprint for adding labels and annotations for image pull secrets, but this also means I may have missed other places they should go. Still, I think this is worth merging to get this formally in as a supported feature.

toddbaert:
> @toddbaert I believe I've resolved all the issues you pointed out. I've largely followed your blueprint for adding labels and annotations for image pull secrets, but this also means I may have missed other places they should go. Still, I think this is worth merging to get this formally in as a supported feature.

Amazing work. I'm reviewing and testing this now.

toddbaert:
Hey @cpitstick-latai ! This all _looks_ good to me, but I tried to install via helm and I get an error (whether or not I pass any values for annotations/labels).

Things I did:

- pulled your branch
- ran `make helm-package`
- installed with `helm upgrade --install open-feature-operator '/home/todd/git/open-feature-operator/charts/open-feature-operator-v0.7.0.tgz'`

Then I got:

```shell
Error: UPGRADE FAILED: parse error at (open-feature-operator/templates/apps_v1_deployment_open-feature-operator-controller-manager.yaml:42): unexpected <range> in parenthesized pipeline
```
It seems like something isn't working with the templating

cpitstick-latai:
Ok I'll take a look and do some thorough testing on my end!

cpitstick-latai:
@toddbaert Alright I've updated it and done much more thorough testing. A couple notes:
- Sometimes when a container is injected, the original deployment has to have `hostNetwork: true` enabled. This doesn't always seem to be true; however, there may be an issue with my cluster.
- Looks like labels, annotations, and imagePullSecrets do not matter for injected pods, as they key off the settings of the original Pod the container is being injected into. So it's only with FlagD standalone and FlagD proxy mode that this matters (and I didn't specifically test those, but I did verify that the labels are being parsed correctly in the main.go function)

cpitstick-latai:
The ugliest/most awkward parts are this line and its companion for annotations: `{{ $labelKeys := keys .Values.labels -}}{{- $labelPairs := list -}}{{- range $key := $labelKeys -}}{{- $labelPairs = append $labelPairs (printf "%s:%s" $key (index $.Values.labels $key)) -}}{{- end -}}{{- join "," $labelPairs }}`

I don't know how to write it better, but it does work and build properly.

cpitstick-latai:
Maybe that nil check should be moved out so that it's always at least an empty map?

toddbaert (approved):
@cpitstick-latai I tested this thoroughly, including flagd-standalone and proxy modes. I did find one very small problem with the annotations if there's no annotations on a flagd deployment (there's a nil exception). I've fixed it with [this commit](https://github.com/open-feature/open-feature-operator/pull/681/commits/952e6288dc96ddca9ef2f4123efcc6ce9193f144); please make sure it makes sense to you and I will merge in the next 24hrs.

cpitstick-latai:
It makes sense. I did merge the if statements a little so that it would make sure the map wasn't nil before the merge check. In this way, it's always an empty map rather than a nil map that's applied to the cluster, and then we don't need nesting!

We have a cyclomatic complexity issue now; however, it seems the additional conditionals tipped this function over the edge.

toddbaert:
> It makes sense. I did merge the if statements a little so that it would make sure the map wasn't nil before the merge check. In this way, it's always an empty map rather than a nil map that's applied to the cluster, and then we don't need nesting!
> 
> We have a cyclomatic complexity issue now; however, it seems the additional conditionals tipped this function over the edge.



> It makes sense. I did merge the if statements a little so that it would make sure the map wasn't nil before the merge check. In this way, it's always an empty map rather than a nil map that's applied to the cluster, and then we don't need nesting!
> 
> We have a cyclomatic complexity issue now; however, it seems the additional conditionals tipped this function over the edge.

I've simplified my solution: I now simply default the annotations on the template with the empty annotation map you created. That fixes the problem without more complexity.

controllers/core/flagd/resources/deployment.go
```diff
@@ -100,9 +109,14 @@ func (r *FlagdDeployment) GetResource(ctx context.Context, flagd *api.Flagd) (cl
...
-	deployment.Spec.Template.Spec.ImagePullSecrets = imagePullSecrets
-
 	// override settings for the injected container for flagd standalone deployment mode
+	deployment.Spec.Template.Spec.ImagePullSecrets = imagePullSecrets
+	if len(r.FlagdConfig.Labels) > 0 {
+		maps.Copy(deployment.Spec.Template.ObjectMeta.Labels, r.FlagdConfig.Labels)
+	}
+	if len(r.FlagdConfig.Annotations) > 0 {
+		maps.Copy(deployment.Spec.Template.ObjectMeta.Annotations, r.FlagdConfig.Annotations)
+	}
```
toddbaert, line 119: Are these lines actually necessary?

I think this is taken care of now by this: https://github.com/open-feature/open-feature-operator/pull/681/files#diff-ea25af66f2087ae4e88440e1dcf34877524a2689948abd770c29116cdaa8c3b1R80-R81

toddbaert: Update: I removed these lines. I think they are redundant. See net changes: https://github.com/open-feature/open-feature-operator/pull/681/commits/59686ced0dfb820fbf5f49b9d28daa65e424eed7

controllers/core/flagd/resources/deployment.go
```diff
@@ -110,14 +111,7 @@ func (r *FlagdDeployment) GetResource(ctx context.Context, flagd *api.Flagd) (cl
 
 	// override settings for the injected container for flagd standalone deployment mode
 	deployment.Spec.Template.Spec.ImagePullSecrets = imagePullSecrets
-	if len(r.FlagdConfig.Labels) > 0 {
-		maps.Copy(deployment.Spec.Template.ObjectMeta.Labels, r.FlagdConfig.Labels)
-	}
-	if len(r.FlagdConfig.Annotations) > 0 {
-		maps.Copy(deployment.Spec.Template.ObjectMeta.Annotations, r.FlagdConfig.Annotations)
-	}
```
toddbaert: With the change above, I don't think we need to do this map copy; we defined `deployment.Spec.Template.ObjectMeta.Annotations` based on the contents of `r.FlagdConfig.Annotation` above.

I tested this locally and the injected pod's annotations look fine.

cpitstick-latai: Yeah, makes sense. I think when I originally coded this I was following a rote algorithm: "Merge labels anywhere there is an image pull secret."

LGTM

cpitstick-latai:
Looks good to me, feel free to merge this at your convenience and when you get some teammates to review it!

beeme1mr (approved):

PR #686, opened 2024-07-03
fix: remove duplicated port in helm

toddbaert:
Delete the default webhook-server port (to avoid duplicate) since we build it from helm


beeme1mr (approved):

config/overlays/helm/exclude-webhook-server-container-port.yaml
```diff
@@ -0,0 +1,17 @@
...
+  template:
+    spec:
+      containers:
+      - name: manager
+        ports:
+        - containerPort: 9443
+          name: webhook-server
+          protocol: TCP
+          $patch: delete
+
```
cpitstick-latai, line 17: Nit: Newline

toddbaert, line 17:
```suggestion
```

Issue #683, opened 2024-06-25
[Helm] Sidecar injection breaks if `defaultNamespace` is set to `null`


cpitstick-latai:
Setting `defaultNamespace` to `null` effectively breaks namespace creation from within the chart. This then breaks sidecar injection.

We need a way to make sure that disabling namespace creation from within the chart (and therefore allowing other tools like Flux to own the namespace creation) does not break functionality of the operator.

cpitstick-latai:
This fixed I believe. The origin of this bug was to disable namespace creation by setting it to something other than the default. THAT issue is now resolved, which means if you set "defaultNamespace: null" you also need to set "namespace.create: false"

Mostly putting here for anyone who needs this for future reference.

PR #678, opened 2024-06-15
feat: Improved custom namespace options

7h3-3mp7y-m4n:
## This PR
- Added the Custom name flexibility
- Add the ability to add additional labels and annotations to the created namespace.

### Related Issues

Fixes #676

PR #702, opened 2024-07-30
feat: Custom labels and annotations for namespace.

cpitstick-latai:
## This PR
Allows custom labels and annotations to be added to a namespace. It also makes it possible to simply disable namespace creation for clusters where ownership of the namespace needs to lie elsewhere for policy or structural reasons.

### Related Issues
Re-implements https://github.com/open-feature/open-feature-operator/pull/678

#### Note 1
Fully rendered `namespace.yaml` after running `make helm-package` (Extracted from `/charts/open-feature-operator-v0.7.1.tgz` using `tar -xvf open-feature-operator-v0.7.1.tgz`)

```
# Only deploy the namespace if the default is being used (helm install should fail if the namespace isnt present)
# when one is defined with -n
{{- if and (eq (include "chart.namespace" .) .Values.defaultNamespace) .Values.namespace.create }}
apiVersion: v1
kind: Namespace
metadata:
  name: '{{ include "chart.namespace" . }}'
  labels:
    control-plane: controller-manager
    {{- range $key, $value := $.Values.namespace.labels }}
    {{ $key }}: {{ $value | quote }}
    {{- end }}
  annotations:
    {{- range $key, $value := $.Values.namespace.annotations }}
    {{ $key }}: {{ $value | quote }}
    {{- end }}
{{ end }}
```
#### Note 2
This does not need to utilize any of the special directives of [strip-kustomize-helm.sh](https://github.com/open-feature/open-feature-operator/blob/main/.github/scripts/strip-kustomize-helm.sh) because it is not referenced in the [helm overlay kustomization](https://github.com/open-feature/open-feature-operator/blob/main/config/overlays/helm/kustomization.yaml)

#### Note 3
I side-stepped the issue that https://github.com/open-feature/open-feature-operator/pull/678 has with the confusion between `namespace.name` and `defaultNamespace` by simply leaving `defaultNamespace` in place and not adding another name field. It is slightly awkward because now we have a `namespace` section and a standalone `defaultNamespace` variable. However, I don't believe it's worth breaking backwards compatibility over this.

### Follow-up Tasks
- Need to validate https://github.com/open-feature/open-feature-operator/issues/683. I believe this PR does fix it.

### How to test
1. Add labels and annotations to the values.
2. Deploy the helm chart with `namespace.create: true` set
3. Validate that the labels and annotations appear on the namespace.
4. Set defaultNamespace to `null` and test with create to be true/false (if false, create namespace externally from the helm chart)


cpitstick-latai:
I have tested adding custom labels and annotations with `create: true` and `defaultNamespace: open-feature-operator-system`, both custom labels and annotations were injected properly.

cpitstick-latai:
I have validated that this fixes https://github.com/open-feature/open-feature-operator/issues/683 if `namespace.create: false` is set

You could say, well "what about if it's null and namespace.create is true?" That scenario doesn't make a ton of sense to me, so I think it's worth skipping it.

Kavindu-Dodan (approved):
looks good from my end :)

beeme1mr (approved):

PR #712, opened 2024-10-14
feat: add flagd-proxy HA configuration

### LINKED ISSUE #609: flagd-proxy HA setup
Based on the behaviour observed and the fixed replica count of `1` ([ref](https://github.com/open-feature/open-feature-operator/blob/04c0a2ca71d16de44a979bb27a025dfdbeb7173c/common/flagdproxy/flagdproxy.go#L138)), I'm wondering if it makes to have two or more instances of the flagd-proxy running in combination with a PodDisruptionBudget.

I can work on this (either as a config option or extending the schema similar to what's currently discussed in https://github.com/open-feature/open-feature-operator/issues/524), but wanted to ask for feedback here first!

xvzf:
## This PR

- Add support for multi-replica flagd-proxy deployments

### Related Issues

Fixes #609

### How to test

Deploy with e.g. `FLAGD_PROXY_REPLICA_COUNT=3`


go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/go-logr/logr v1.4.2
 	github.com/golang/mock v1.4.4
 	github.com/kelseyhightower/envconfig v1.4.0
-	github.com/open-feature/open-feature-operator/apis v0.2.41-0.20240506125212-c4831a3cdc00
+	github.com/open-feature/open-feature-operator/apis v0.2.44
```
xvzf, line 9: With the orginal go.mod, the app doesn't compile (tested on 3 different machines with an empty gomod cache)

beeme1mr:
Thanks for the PR! I'll look into the image scanning CI issue, which is unrelated to this change.

common/flagdproxy/flagdproxy.go
```diff
@@ -71,58 +76,99 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
...
+		// If the object exists but is not managed by OFO, return an error
+		if !notFound && !common.IsManagedByOFO(old) {
+			return fmt.Errorf("%s not managed by OFO", obj.GetName())
+		}
+
+		// If the object is not found, we will create it
+		if notFound {
+			return f.Client.Create(ctx, obj)
+		}
+
```
thisthat, line 122: we can simplify a bit the logic here swapping the order
```suggestion
                 // If the object is not found, we will create it
		if notFound {
			return f.Client.Create(ctx, obj)
		}
		// If the object exists but is not managed by OFO, return an error
		if !common.IsManagedByOFO(old) {
			return fmt.Errorf("%s not managed by OFO", obj.GetName())
		}
```

common/flagdproxy/flagdproxy.go
```diff
@@ -71,58 +76,99 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
...
+		// Try to get the existing object
+		err := f.Client.Get(ctx, client.ObjectKey{Name: old.GetName(), Namespace: old.GetNamespace()}, old)
+		notFound := errors.IsNotFound(err)
+		if err != nil && !notFound {
+			return err
+		}
+
+		// If the object exists but is not managed by OFO, return an error
+		if !notFound && !common.IsManagedByOFO(old) {
+			return fmt.Errorf("%s not managed by OFO", obj.GetName())
```
thisthat, line 115: I think here we should return `nil` and log the message.
If there exists a proxy not managed by OFO, I don't think we should retry continuously here.

common/flagdproxy/flagdproxy.go
```diff
@@ -71,58 +76,99 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
...
-	return f.Client.Update(ctx, obj)
+// ensureFlagdProxyResource ensures that the given object is reconciled in the cluster. If the object does not exist, it will be created.
+func (f *FlagdProxyHandler) ensureFlagdProxyResource(ctx context.Context, obj client.Object) error {
+	if obj == nil {
+		return fmt.Errorf("object is nil")
+	}
+
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var old = obj.DeepCopyObject().(client.Object)
+
```
thisthat, line 105: Since this is called potentially multiple times, maybe it is worth keeping the info logs to support debug activities.

thisthat (commented):
Thank you @xvzf for the PR with some extra refactoring  🙌 
I left only some minor comments. In particular, I am not 100% sure if we should retry continuously to reconcile on an unmanaged OFO resource or we should skip it. What's your opinion here?

common/flagdproxy/flagdproxy.go
xvzf: nice catch!

common/flagdproxy/flagdproxy.go
xvzf: I've been unsure here. I have the feeling these components are **supposed** to be managed by OFO, so I think it should be an error and it should retry here (though with an exponential backoff). Wdyt?

beeme1mr:
Hey @xvzf, could you please resolve any open threads that have been resolved? I'll rereview once you're ready. The change looks great so far. Thanks!

xvzf:
Here we go @beeme1mr @thisthat - been off for a couple of days, sorry for the delay

beeme1mr (approved):

thisthat (approved):
Thanks a lot for the contribution @xvzf 🙌

beeme1mr:
The change is now live in v0.8.0. Thanks @xvzf!

xvzf:
Thanks for the swift release @beeme1mr!

beeme1mr:
Hey @xvzf, it looks like this change introduced new permission requirements. Would you be able to update the RBAC configuration?

https://github.com/open-feature/open-feature-operator/issues/717

common/flagdproxy/flagdproxy.go
```diff
@@ -71,58 +76,99 @@ func (f *FlagdProxyHandler) Config() *FlagdProxyConfiguration {
...
+			Kind:       "Service",
+			APIVersion: "v1",
+		},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:            FlagdProxyServiceName,
 			Namespace:       f.config.Namespace,
 			OwnerReferences: []metav1.OwnerReference{*ownerReference},
+			Labels: map[string]string{
+				common.ManagedByAnnotationKey: common.ManagedByAnnotationValue,
+			},
```
wondertroy, line 171: I just noticed when upgrading to open-feature-operator that previous older flagdProxy resources that already existed did not have this label - and therefore ran in to the 'not managed by OFO' in the ensureFlagdProxyResource method. 

It was an easy workaround to manually add this label to the older resources but worth mentioning

wondertroy: The previous resources did seem to have the `common.ManagedByAnnotationKey: common.ManagedByAnnotationValue,` in the Selector though

xvzf:
Yes, gonna have a look this morning!

PR #715, opened 2024-10-29
feat: allow setting mutating webhook failurepolicy and objectSelector via helm

### LINKED ISSUE #654: Sidecar is not injected during Kubernetes Cluster startup
Hi Team,

We're excited about integrating the open feature into our current project. Every day, we restart our whole Kubernetes cluster. However, during startup, the open feature operator fails to inject the flagD sidecar into the respective components. Our assumption was that all services would come up simultaneously during startup. It seems that the open feature isn't healthy enough to inject the sidecar. Do you have any ideas on how to fix this?

Oro:
## This PR

- Allow setting mutatingWebhook failurePolicy and objectSelector via helm

This makes it possible to configure new pods to wait until the webhook is available. Otherwise pods will start and not get mutated, i.e. will be missing the flagd sidecar.
The objectSelector is needed so that the webhook itself can still start up as otherwise it would fail while waiting on itself.

### Related Issues

Fixes #654


beeme1mr (approved):
Looks good, thanks.

thisthat (approved):

thisthat:
Thanks @Oro for the PR 🙌 
Could you please run `./.github/scripts/generate-helm-docs.sh` to re-generate the Helm docs with these new options?

lukas-reining (approved):

PR #721, opened 2024-10-30
ci(trivy): fetch vulnerabilities DB from ERC

beeme1mr:
We're running into GHCR throttling issues. Switching to ECR should resolve this issue.

https://github.com/aquasecurity/trivy-action/issues/389#issuecomment-2373539982

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -83,6 +83,7 @@ jobs:
           output: "trivy-results.sarif"
           severity: "CRITICAL,HIGH"
         env:
+          # use an alternative trivvy db to avoid rate limits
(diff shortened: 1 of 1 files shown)


toddbaert (approved):

.github/workflows/pr-checks.yml
```diff
@@ -82,6 +82,8 @@ jobs:
           template: "@/contrib/sarif.tpl"
           output: "trivy-results.sarif"
           severity: "CRITICAL,HIGH"
+        env:
+          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2
```
toddbaert, line 86:
```suggestion
          # use an alternative trivvy db to avoid rate limits
          TRIVY_DB_REPOSITORY: public.ecr.aws/aquasecurity/trivy-db:2
```

PR #725, opened 2025-01-10
feat: add gateway api

### LINKED ISSUE #723: Add gateway API routes to flagd CRD
#615 with the implementation in #633 brought in a new CRD for deploying a standalone flagd with service and optional ingress. 
I would like to have the option to deploy matching [Gateway API routes](https://github.com/kubernetes-sigs/gateway-api/blob/main/apis/v1/httproute_types.go) instead of an ingress with the increasing adoption of Gateway API. 

I will open a PR the next days if no arguments against that come up before :) 

cc @open-feature/cloud-native-maintainers @toddbaert @beeme1mr

lukas-reining:
## This PR

Adds Gateway API support as described in #723, similar to the ingress support in #633.

The operator only creates an `HTTPRoute` for all endpoints instead of explicitly creating a `GRPCRoute` for the GRPC 
endpoints, because we are using GRPC Gateway to enable HTTP+JSON for the GRPC endpoints. 
This means that these endpoint not only support GRPC, but also plain HTTP. Because of this, `GRPCRoute` does not work 
well for these endpoints.

I decided to not make the paths/methods configurable, while they are for ingress, as they can not be adjusted in flagd anyways.

The Go version had to be updated as the gateway api package is not working with Go 1.21. Due to this I also had to update the chainsaw version. 

### Related Issues

Fixes #723


config/rbac/role.yaml
beeme1mr, line 1: Why are some many permissions changed in this file?

beeme1mr (approved):

toddbaert:
This is really cool.

PR #797, opened 2026-02-04
fix: Set appProtocol for service resources

### LINKED ISSUE #796: Set appProtocol for service resources
Hi,

We are exposing a `flagd` instance via a GRPCRoute. When attempting to retrieve feature flag details, requests fail with `StatusCode.UNKNOWN`. After some investigation, we found that explicitly setting `appProtocol: grpc` on the `flagd` service port resolves the issue and allows requests to succeed as expected.

The primary benefit of specifying the `appProtocol` field is that the protocol is no longer inferred, which reduces ambiguity and prevents related issues. It also makes the manifests more explicit and future-proof, while serving as clearer documentation for users.

That seem to happen because we use Istio, where the gateway will forward all incoming HTTP requests using HTTP/1.1. https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

I'm more than happy to raise a PR and resolve it.

vspallas-flyr:
## This PR

- adds the `appProtocol` in the service resource for all the exposed ports

### Related Issues

- #796

### DIFF
--- a/internal/common/flagdproxy/flagdproxy.go
+++ b/internal/common/flagdproxy/flagdproxy.go
@@ -16,6 +16,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/util/retry"
+	"k8s.io/utils/ptr"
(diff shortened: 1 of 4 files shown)


beeme1mr (approved):

toddbaert (approved):
Nice! Thanks @vspallas-flyr , sorry about the wait.

PR #836, opened 2026-05-02
chore: migrate Renovate config preset

alceops:
## Summary
- migrate Renovate from `config:base` to `config:recommended`
- migrate the regex manager block to Renovate custom managers while preserving the existing match/file patterns

## Verification
- `python3 -m json.tool renovate.json`
- `git diff --check`

No runtime or cluster tests were run; this is a Renovate config-only change.


beeme1mr (approved):
Thanks @alceops

PR #853, opened 2026-07-30
feat: expose pod scheduling fields on the Flagd CRD

### LINKED ISSUE #849: Expose pod scheduling fields (nodeSelector/affinity/tolerations) on the Flagd CRD
## What

`Flagd.spec` (`core.openfeature.dev/v1beta1`) has no pod scheduling knobs. The operator builds the flagd Deployment's pod template internally, so there's no way to influence where the pods land.

Current spec surface: `featureFlagSource`, `replicas`, `serviceType`, `serviceAccountName`, `ingress`, `gatewayApiRoutes`. None of `nodeSelector`, `affinity`, `tolerations`, or `topologySpreadConstraints`.

## Why

Near term: I want to pin flagd to a dedicated node for isolated load/stress testing, keeping its CPU off the nodes running the consumers.

Longer term: isolate flagd onto its own nodepool or instance types (dedicated nodepool via taints + tolerations, spread across instances via topologySpreadConstraints) so flagd capacity is managed independently of the app workloads.

Both are standard k8s placement, but the CRD gives me no handle on it. Today the options are dropping the CRD for a hand-rolled Deployment, or node taints. Both throw away the operator reconciliation that's the reason to use the CRD in the first place.

## Ask

Plumb scheduling fields from `Flagd.spec` into the generated Deployment's pod template. Either:

- discrete fields: `nodeSelector`, `affinity`, `tolerations`, `topologySpreadConstraints`, or
- a single `pod`/`podTemplate` overrides block if you'd rather not enumerate.

Discrete fields keep the surface predictable.

The pod template is built in `GetResource` at `internal/controller/core/flagd/resources/deployment.go`, where the `PodSpec` literal currently only sets `ServiceAccountName`. The new fields would be read from the CR and set on that same literal.

Happy to PR if there's agreement on the shape.

readyon-emerson:
## This PR

Adds `nodeSelector`, `affinity`, `tolerations` and `topologySpreadConstraints` to `FlagdSpec` and passes them through to the pod template of the generated `Deployment`.

All four fields are optional and default to unset, so existing `Flagd` resources produce an identical `Deployment` to before. Discrete fields rather than a nested `podTemplate`, as agreed in the issue.

- `api/core/v1beta1/flagd_types.go`: new optional fields on `FlagdSpec`
- `internal/controller/core/flagd/resources/deployment.go`: fields set on the `PodSpec` of the pod template
- regenerated `zz_generated.deepcopy.go`, `config/crd/bases/core.openfeature.dev_flagds.yaml` and `docs/crds.md`
- `docs/flagd.md`: new "Pod scheduling" section with an example

### Related Issues

Fixes #849

### Notes

`Flagd` is single-version (`v1beta1`), and these are additive optional fields, so no conversion work is needed.

### Follow-up Tasks

None.

### How to test

Unit tests cover both the populated and the unset case:

```
make unit-test
```

A chainsaw case is added at `test/e2e/chainsaw/flagd-with-pod-scheduling` that applies a `Flagd` with all four fields set and asserts they land on the generated `Deployment`'s pod template, with both replicas becoming ready:

```
make e2e-test-chainsaw
```

The scheduling values in the e2e case are deliberately satisfiable on the kind cluster (`kubernetes.io/os: linux`, `whenUnsatisfiable: ScheduleAnyway`) so the assertion tests the plumbing rather than the scheduler.

### DIFF
--- a/api/core/v1beta1/flagd_types.go
+++ b/api/core/v1beta1/flagd_types.go
@@ -47,6 +47,23 @@ type FlagdSpec struct {
 	// the feature flag configurations
 	FeatureFlagSource string `json:"featureFlagSource"`
 
+	// NodeSelector defines the node labels a node must match for the flagd pods to be scheduled on it
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Affinity defines the scheduling constraints of the flagd pods
+	// +optional
+	Affinity *v1.Affinity `json:"affinity,omitempty"`
+
+	// Tolerations defines the tolerations of the flagd pods, allowing them to be scheduled onto
+	// nodes with matching taints
+	// +optional
+	Tolerations []v1.Toleration `json:"tolerations,omitempty"`
+
+	// TopologySpreadConstraints defines how the flagd pods are spread across topology domains
+	// +optional
+	TopologySpreadConstraints []v1.TopologySpreadConstraint `json:"topologySpreadConstraints,omitempty"`
+
 	// Ingress
 	// +optional
 	Ingress IngressSpec `json:"ingress"`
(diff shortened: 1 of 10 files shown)


readyon-emerson:
@toddbaert tagging you because I can't add you as a reviewer.

Sorry for the delay on opening the PR, I didn't notice your answer.

PR #855, opened 2026-08-26
fix: drop ALL capabilities in controller-manager

lukas-loibnegger-dt:
## This PR

drop all capabilities for the controler-manager


toddbaert (approved):
Nice catch - this should be automatically picked up by out helm tooling and apply there as well.

toddbaert:
Thanks @lukas-loibnegger-dt !

